PR #9, opened 2026-08-05
Enforce LF line endings, and keep git blame readable across it

JerryHyun:
## ⚠️ Please merge with a merge commit — not squash, not rebase

The second commit's `.git-blame-ignore-revs` names the first commit's SHA. A squash or rebase merge gives that commit a new SHA, at which point the file **silently stops doing anything** — git ignores an absent SHA without a warning. A merge commit preserves both SHAs. (If squashing is house style, the entry just needs re-pointing afterwards; happy to follow up.)

## Commit 1 — enforce LF

The repo had **mixed** line endings committed: most files LF, but 25 Rust files under `src-tauri/` carried CRLF in their blobs (`workdir.rs` alone had 2,683 CRs). Nothing declared a policy, so which of the two a file got came down to whichever `core.autocrlf` the last committer's machine had.

`* text=auto eol=lf` pins both sides: LF in the repository **and in the working tree**. The working-tree half matters concretely on Windows, where `core.autocrlf=true` checks files out as CRLF — any tool that rewrites a source file with LF endings then leaves it showing as modified with an *empty* `git diff`. Several generators in this repo can produce exactly that state.

Every binary asset type currently committed is additionally marked `binary`, taken from what `git ls-files` actually lists (`.webp .png .jpg .ico .icns .mp4 .webm`), not from memory. `.svg` is deliberately absent — it's XML, and normalizing it is correct.

The 25 renormalized files are verified **pure CR-removal**: stripping CRs from each old blob hashes identically to the new blob. Not one byte of content moved, and `cargo build --locked` + `cargo build --examples` pass on the result.

## Commit 2 — keep blame readable

A whole-file rewrite is non-semantic, but `git blame` doesn't know that: without countermeasures, all 2,683 lines of `workdir.rs` (and everything in `submodule.rs`, `tool_settings.rs`, and 22 others) get attributed to the normalization commit. CONTRIBUTING.md opens by teaching history archaeology as the way to get oriented here, so blanking blame on the largest files in the repo is a real cost.

`.git-blame-ignore-revs` names the normalization commit. GitHub applies it to its own blame view with no setup; locally it's one opt-in per clone, now documented in CONTRIBUTING.md:

```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

Measured on `workdir.rs`: without the file, Jerry Hyun 2683/2683; with it, Jiucheng Zang 2527 / Jerry Hyun 156. Not a full restoration (~6% of lines can't be mapped back through a whole-file rewrite — a limitation of `--ignore-rev` itself), but 94% back is the difference between blame being useful and useless.

CONTRIBUTING.md also warns about the one sharp edge, verified on git 2.53.0: the config is repo-wide but the file only exists on commits that have it, so checking out anything older (a tag, a bisect step) makes every `git blame` in that worktree **fatal** rather than fall back — with `git config --unset blame.ignoreRevsFile` as the way back out.

## Verification

- Pure CR-removal proven blob-by-blob (hash comparison).
- `cargo build --locked` and `cargo build --examples` pass on Windows.
- Full frontend suite unaffected (vitest / svelte-check / Playwright all green).
- Blame restoration re-measured after the final rebase (2527/2683).


zangjiucheng:
Thanks for the PR; that's a good catch because I develop on both platforms, but I will format this on my end and apply your suggestions. This is outside my scope for review, sorry.

PR #11, opened 2026-08-05
Fix: uncaught TypeError at boot on every launch

JerryHyun:
## What happens

`src/main.ts` primes the submodule-navigator strip once at mount:

```ts
submoduleNavCtrl.refresh(bridge.CUR_REPO as unknown as string);
```

At mount, `CUR_REPO` is still `null` — `legacy/main.ts` starts `openRepo()` without awaiting it, and `openRepo` itself awaits (bisect cancellation, then the graph stream) before it ever assigns `CUR_REPO = path`. So the module body always finishes, and the mounts always run, while `CUR_REPO` is null — on a cold boot **and** on a deep-link window opened straight onto a repo.

`refresh()` declared its parameter as `string` and handed it straight to `basename()`:

```
TypeError: Cannot read properties of null (reading 'replace')
    at basename (submodulenav.svelte.ts:22)
    at SubmoduleNavState.refresh (submodulenav.svelte.ts:132)
    at src/main.ts
```

Because `refresh` is async and the call isn't awaited, this surfaced as an unhandled rejection rather than a hard failure — the app still came up, which is why it went unnoticed. But the strip was left unprimed and the console carried an exception on every launch.

## The fix

At the callee, not a crusade against the call sites: `refresh(repo: string | null)` resets to the empty strip when there is no repo — the same state `reset()` already produces when a repo is closed. The mount-time call site now hands the null over honestly instead of casting it away.

Worth being precise: the cast is **not** the defect. `CUR_REPO` lives in `@ts-nocheck` legacy code as `export let CUR_REPO = null`, so `as unknown as string` is the established idiom for reading it — 24 sites in `src/main.ts` before this change. The others are safe for their own reasons (behind a menu action/shortcut that can't fire before a repo is open, behind a watcher/poll callback that null-checks first, or calling a callee that guards). This one was the only one that runs at mount.

The null guard sits **below** the `!IN_TAURI` design-mode branch on purpose: design mode has no repo for its whole session, so a guard placed first would blank the browser preview's demo strip.

## Verification

- Reproduced in a real browser through the e2e Tauri mock: one boot-time pageerror on `dev`, none after the fix.
- 5 regression tests: `null`, empty string, clearing an already-populated strip, and the guard staying below the design-mode branch. The first four fail against the old signature with exactly this TypeError.
- `svelte-check` 0 errors; full vitest suite passes.

### DIFF
--- a/src/islands/submodulenav/submodulenav.svelte.test.ts
+++ b/src/islands/submodulenav/submodulenav.svelte.test.ts
@@ -260,3 +260,52 @@ describe("horizontalWheelDelta — wheel-to-scroll on the overflowing strip", ()
     expect(horizontalWheelDelta({ deltaX: 0, deltaY: 3, deltaMode: 1, ...overflow })).toBe(48);
   });
 });
+
+describe("refresh with no repo open", () => {
+  beforeEach(() => {
+    mockInTauri = true;
+    mockNavStack = [];
+    mockCurRepo = "";
+    submoduleNavCtrl.reset();
+  });
+
+  // REGRESSION: src/main.ts calls this once at boot, with whatever
+  // `bridge.CUR_REPO` currently is — and at boot with no repo open that is
+  // `null`. `refresh` used to take it as `string` (the call site casts through
+  // `as unknown as string`) and hand it straight to `basename()`, which does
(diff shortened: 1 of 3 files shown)


zangjiucheng (approved):
LGTM, test on my local end can reproduce this issue and resolved by the patch.

PR #12, opened 2026-08-05
Fix: branch-visibility changes never reached the graph

JerryHyun:
## Symptom

Toggling a branch checkbox — or clicking Auto / Show all branches / Hide all branches — visibly changed the sidebar's own state and persisted correctly, and did nothing at all to the commit graph. Every commit stayed, every hidden branch kept its label, until some unrelated event forced a full reload (the manual Refresh button, or reopening the repo).

Measured on a 93k-commit repo with "Hide all branches" persisted as `{local: [], remote: []}`:

| layer | result |
| --- | --- |
| `walk_repo`, no filter | 93,545 commits / 19 branch chips / 392 remote chips |
| `walk_repo`, filtered | 58,630 commits / 0 chips |
| actually on screen | 93,544 commits, all chips present |

The backend was correct at every layer — the registry lookup resolved the row, `walk_repo` honored the filter, `filter_hidden_chips` dropped the chips. The change was being discarded in the frontend's reload path.

## Root cause

`persistVisibleBranches` called `bridge.reloadGraph(true)`. That signature is `reloadGraph(preserveRow, forceFull = false)`, so this only asked to preserve the selected row and left the fast path enabled. But a visibility change decides which commits the walk seeds from, so it **adds or removes rows** — and the fast path only ever remaps ref chips over rows that are already loaded. It cannot express the change; and because a visibility toggle moves no ref and does not move HEAD, that path classified it as a pure working-tree change and returned successfully having done nothing.

## Fix

- `persistVisibleBranches` now forces a full reload (`reloadGraph(true, true)`), rather than teaching the fast path filter-awareness — no amount of better change detection helps when the path has no mechanism for adding or removing rows in the first place.
- The flip side: a full reload is expensive and ends by re-reading the persisted filter, so the `set_visible_branches` **Result is now checked** before paying for one. Previously only the thrown half was handled; a rejected Result went straight on to force a re-walk that could only put the old picture back, with the user's checkbox silently snapping back and nothing said about why. Both failure halves now warn through Tama and re-read the stored filter.
- Comment repairs that this change made necessary: `tryFastRefresh`'s note on the auto-visibility pass described Gate B promoting it to a full reload, which stops being how it works with this fix; and the branch-rename path now states why it deliberately stays on the fast path (a rename moves no commit, so no row can appear or disappear).

## Verification

- Regression tests for all four entry points (toggle / show all / hide all / auto), asserting both the `(true, true)` call and the **absence** of the fast-path-eligible one-argument form. All four fail against the old call.
- 2 tests for the failed-persist path (rejected Result, thrown error): no reload, a warning, and a re-read of the stored filter.
- 215 sidebar tests / full suite green; `svelte-check` 0 errors.

### DIFF
--- a/src/islands/sidebar/sidebar.svelte.test.ts
+++ b/src/islands/sidebar/sidebar.svelte.test.ts
@@ -119,6 +119,10 @@ function resetAll() {
   // care about this" reasoning as submoduleStatus above — only the
   // "branch visibility" describe block overrides this per-test.
   vi.mocked(commands.getVisibleBranches).mockResolvedValue(ok({ local: null, remote: null, auto: false }));
+  // Default: persisting succeeds. persistVisibleBranches checks this Result and
+  // skips the graph reload when it failed, so without a default every test that
+  // changes visibility would take the failure path instead.
+  vi.mocked(commands.setVisibleBranches).mockResolvedValue(ok(null));
 }
 
 beforeEach(() => {
@@ -246,7 +250,8 @@ describe("branch visibility", () => {
     await sidebarCtrl.toggleBranchVisible("/repo", "local", "dev");
     expect(sidebarCtrl.visibleLocal).toEqual(["main"]);
     expect(commands.setVisibleBranches).toHaveBeenCalledWith("/repo", false, ["main"], null);
-    expect(bridge.reloadGraph).toHaveBeenCalledWith(true);
+    // forceFull — see persistVisibleBranches' own doc comment.
+    expect(bridge.reloadGraph).toHaveBeenCalledWith(true, true);
   });
 
   it("toggleBranchVisible: a manual toggle turns auto mode off (grabbing the wheel exits autopilot)", async () => {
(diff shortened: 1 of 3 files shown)


zangjiucheng (approved):
Thanks for the thorough writeup, the measured table made the root cause obvious. The diagnosis holds up: reloadGraph(true) left the fast path enabled, and that path only remaps ref chips over rows that are already loaded, so a visibility change that adds or removes rows had no way to land. Forcing the full reload is the right lever here, rather than trying to teach the fast path to be filter-aware.

The failure handling is a real improvement on its own. Checking the Result and re-reading the stored filter (which restores auto/local/remote together) means a failed save snaps the checkboxes back to the truth instead of paying for a walk that arrives at the same picture. Keeping rename and new-branch on the fast path is also correct, since neither moves a commit.

Tests cover all four entry points and both failure modes, and the default success mock in resetAll is a nice touch so the other visibility tests do not drift onto the failure path.

One tiny follow-up, not for this PR: the new "Couldn't save which branches to show." string is hardcoded English. That is fine on dev, and I will wrap it in a t() key when the i18n branch lands, so nothing needed from you here.

Approving. Thanks for the careful fix.

PR #13, opened 2026-08-05
Sidebar: group branches, remotes, and tags into a folder tree

JerryHyun:
## What

`/`-segmented refs now render as a collapsible folder tree (the Git Fork / Tower convention) instead of a flat list — a repo with a dozen `feature/*` branches shows one `feature` folder, not a wall of rows. Pure frontend; no backend or IPC changes.

**Ordering.** At every level: folders A→Z first, then plain refs A→Z. The comparison is `localeCompare(…, { numeric: true, sensitivity: "base" })`, so `release/1.9` sorts before `release/1.10`.

**Folders start closed — with two exceptions.**
- The folder path containing **HEAD** opens by default, so "where am I" survives the decluttering.
- A **remote's own root node** opens by default. The Remotes section is already a `<details>` that stays shut until clicked; leaving each remote shut too would mean two clicks before a single branch appears. Folders *inside* a remote get the normal closed default.

An explicit click always wins over both defaults. Collapse state is keyed per section and per full path (`origin/feat` and `upstream/feat` collapse independently), and is reset when the repo changes.

**Remotes are tree nodes.** A remote branch belongs *under* its remote — `origin` is a depth-0 node (cloud icon), its branches and folders nest beneath it. Grouping keys are full ref names, so same-named folders under different remotes can't collide.

**Tags group too** (`v1.0/rc1`, `v1.0/rc2` → a `v1.0` folder), with the same rules.

**Filter interaction.** While the ref filter has text, the tree force-expands (a match hidden inside a closed folder would look like a miss) and the collapse-all buttons hide; clearing the filter restores the previous collapse state. Each section header gets a collapse-all/expand-all toggle, shown only when that section actually has folders.

**Rows.** Folder rows are keyboard-accessible (`role="button"`, `tabindex`, `aria-expanded`) and show a count badge while collapsed. Leaf rows keep their full ref name in `data-branch`/`data-tag`, so every existing context-menu/checkout/visibility path is untouched.

## Implementation notes

- `buildRefRows` is a pure function (items + collapse predicate → ordered row list) in `sidebar.svelte.ts`; the view renders one `{#each}` per section over its output. Indentation is a CSS `--depth` variable, not nested markup.
- Each section's folder-path list is derived once per ref-list change rather than recomputed in the summary markup on every render (relevant with several hundred remote names).
- Demo refs extended (`release/1.9`/`1.10`, `upstream/feat/inline-diff`, `v1.0/rc1`/`rc2`) so design mode exercises numeric ordering, same-named folders under two remotes, and tag folders.

## Coverage

- `sidebar.svelte.test.ts` covers `buildRefRows` and the collapse controller exhaustively (ordering, defaults, per-remote independence, repo-switch reset).
- New `e2e/ref-folder-tree.spec.ts` covers the **view** wiring in a real browser (design mode): rendered row order, collapse keys, indentation depth, filter force-expand/restore, collapse-all scope. This is deliberately the *only* spec that knows refs are grouped — `open-repo.spec.ts` asserts the branch count and `main`, neither of which depends on the shape.
- Full suite: vitest 45 files green, `svelte-check` 0 errors, Playwright specs pass.

3-filtered
1-default-closed
2-expanded


zangjiucheng (changes requested):
- Blocking: feature/widget is created with git branch and no checkout, so HEAD stays on main, its feature folder defaults collapsed, and a collapsed folder renders no children, so this assert times out. Expand the folder first (or don't collapse single-leaf folders), and add this spec to CI.
https://github.com/zangjiucheng/GitCat/blob/4699c7689ef92b549a9f2c0433b3fb76c7b096db/e2e/open-repo.spec.ts#L22
- Missing test: the repo-switch reset has no test that goes through refresh, so deleting the guard stays green. Add one that opens a folder, switches repo, and asserts it cleared.
https://github.com/zangjiucheng/GitCat/blob/4699c7689ef92b549a9f2c0433b3fb76c7b096db/src/islands/sidebar/sidebar.svelte.ts#L562-L564
- Merge: branched off dev before 8a73662, so this Local <details> conflicts. Keep open plus your new summary markup.
https://github.com/zangjiucheng/GitCat/blob/4699c7689ef92b549a9f2c0433b3fb76c7b096db/src/islands/sidebar/Sidebar.svelte#L264
- Same merge: make the e2e helper open-idempotent, since it clicks Local's summary open, which collapses it once Local defaults to open.
https://github.com/zangjiucheng/GitCat/blob/4699c7689ef92b549a9f2c0433b3fb76c7b096db/e2e/ref-folder-tree.spec.ts#L50-L53
- Nit: this Space handler toggles but misses preventDefault(), so Space also scrolls the ref list.
https://github.com/zangjiucheng/GitCat/blob/4699c7689ef92b549a9f2c0433b3fb76c7b096db/src/islands/sidebar/Sidebar.svelte#L294

JerryHyun:
All five addressed in 3d0e3eb (on top of a merge of current `dev`). Point by point:

**Blocking — `open-repo.spec.ts`:** now opens the `feature` folder before asserting its child, exactly as you described — closed folder renders no children, and `repo.branch()` doesn't check out so the HEAD-path exception doesn't apply. Two things worth knowing from making it actually pass:
- The spec was failing **before** the folder assert on this branch: `.repo-pick` opens the repositories dashboard now (not a folder picker), and the Tauri mock lacked the dashboard's commands plus the six `openRepo` fans out to (`conflict_status`, `get_visible_branches`, `submodule_superproject_chain`, `bisect_status`, `claim_repo_summary_first_open`, `run_hooks`) — every caller catches and `console.error`s, so the gaps never failed a test. This PR now carries that mock repair; it's byte-identical to the one in #14, so whichever merges second is clean on that file.
- **CI**: #14 wires this whole suite into ci.yml as its own parallel job (browser cache, failure-only report artifact). Kept there rather than duplicated here — happy to fold it into this PR instead if you'd prefer.
- I stayed with "expand first" over "don't collapse single-leaf folders": a folder that's open or closed depending on how many children it currently has changes shape as branches come and go, which felt less predictable than one consistent default. Can switch if you'd rather have the other.

**Repo-switch reset test:** two tests now drive it through `refresh()` — open a folder in repo A, refresh repo B, assert cleared; and re-refresh the *same* repo, assert the fold survives (every mutation ends in a same-repo refresh, so a checkout must not snap the tree shut). Verified by mutation: deleting the guard fails exactly the first; clearing unconditionally would fail the second.

**Merge:** resolved as you said — Local `<details>` keeps its new `open` default plus this branch's summary markup.

**Helper idempotence:** it now reads `details.open` and only clicks sections that are actually shut, so it survives any future changes to which sections default open.

**Space:** `preventDefault()` added on the folder-row handler in all three sections.

`svelte-check` 0 errors · 250 sidebar unit tests · all 12 Playwright specs green on the merged branch.

zangjiucheng (approved):
Thanks for the quick turnaround. Every review point is addressed and it looks good:

- `open-repo.spec.ts` opens the `feature` folder before asserting the nested branch, and the new `tauriMock.ts` fixture fixes the harness (it was failing earlier at the repo-open step).
- The repo-switch reset has real coverage now, both through `refresh()`: cross-repo clears, same-repo preserves.
- The dev merge is clean and the Local `<details open>` is preserved (line 264); `openSidebar` is now open-idempotent.
- The folder-row Space handlers call `preventDefault()`.

Ran the sidebar unit tests locally, 250 pass. Approving.

Trivial and non-blocking: the remote/tag `onkeydown` `if` lines are indented a little oddly; prettier will flatten them.

PR #14, opened 2026-08-05
CI: run the Playwright suite (and repair it first)

JerryHyun:
## Why

The `e2e/` suite has never run anywhere but a developer's own machine. It was clearly built for CI — `playwright.config.ts` branches on `process.env.CI` for retries, `forbidOnly`, and `reuseExistingServer` — but no workflow invoked it, so `pnpm test:e2e` was the one test command nothing enforced. Predictably, it had rotted.

## Repairing the suite

`open-repo.spec.ts` had gone stale in three independent ways (first one reproduced on `dev` before touching anything):

1. **The entry point moved.** The spec clicks `.repo-pick` expecting a folder picker; that chip now opens the repositories dashboard (`legacy/main.ts`'s own handler), and picking a folder lives behind the dashboard's "+ Add repository…". The spec failed on its first assertion — the chip still read "Open a repository…". It now drives the real path (chip → dashboard → add).
2. **The mock was missing the commands that path uses.** Added: the dialog plugin's `plugin:dialog|open` (islands go through `invoke`, unlike legacy's raw `window.__TAURI__.dialog.open`, which was already stubbed) and an in-memory tracked-repo list that deliberately starts **empty**, so a test has to pick a folder like a user does. Then the six commands `openRepo` fans out to (`conflict_status`, `get_visible_branches`, `submodule_superproject_chain`, `bisect_status`, `claim_repo_summary_first_open`, `run_hooks`) — every caller catches and `console.error`s, so the gaps never failed a test; they just left the UI in a state no real repo produces. They now answer "nothing to report"; `claim_repo_summary_first_open` returns `false` specifically so the one-time Repository Summary modal doesn't cover the app on the first open of every fixture repo (each one being, necessarily, a first open).
3. **Sidebar sections are `<details>` with no `open` attribute** — rows render into the DOM but stay hidden until the group is expanded, so the spec's `toBeVisible` assertions could never have passed no matter what `list_refs` returned. The spec now expands the Local group first, then asserts `#cntLocal` (the shape-independent half of "list_refs was read") plus `main` visible.

## The CI job

- Its own job, parallel to `rust` and `frontend`: it needs a browser download those steps have no use for, and a flake there shouldn't sit on already-finished type-check/unit-test results.
- Browsers cached against `pnpm-lock.yaml` (what pins `@playwright/test`; a browser build only works with the driver it shipped with). `--with-deps` still runs on a cache hit — the system libraries it installs aren't in the cached path.
- CI reporter is now `github` + `html` (was `github` only), so the HTML report exists to upload as an artifact **on failure only** — that's where `trace: "on-first-retry"` becomes readable.
- README's command block gains the `pnpm test:e2e` line.

## Verification

Locally, including with `CI=true` so the config's CI branches are the ones under test: all specs pass, GitHub annotations emitted, `playwright-report/` written. Also verified merged together with the other open branches (including the sidebar folder tree, where `feature/widget` becomes a child of a closed folder) — assertions still hold.

**Heads-up:** this adds a `pnpm/action-setup` step with `version: 10`, matching the four existing ones. #15 removes that input everywhere in favour of `package.json`'s `packageManager` — whichever of the two lands second needs this one line brought in line with it.


JerryHyun:
Merged current dev: took #13's open-repo.spec.ts wholesale (it carries both the dashboard-flow repair and the folder handling), and removed the new e2e job's `version: 10` input now that #15's `packageManager` field is on dev — the one-line reconciliation both PR descriptions flagged. 12/12 specs green on the merged tree.

zangjiucheng (approved):
Verified the repaired spec and full mocks are already on dev and the config is sound (CI starts its own vite via webServer, and the no-version pnpm setup matches the other jobs), so this runs the suite green; approving.

PR #18, opened 2026-08-06
Graph: inline ref labels (default) + local/remote chip distinction

JerryHyun:
Hi! This one changes how branch/tag labels appear on the commit graph, so let me start with the part you'll care about most.

## The big change (and the escape hatch)

**Ref labels now sit inline, right before the commit subject, and the lanes start at the far left** — instead of the dedicated left column. This is the new default, so existing users will see the change on upgrade.

I know the left column is a feature you built recently, so it's fully kept: there's a new select in `Settings → Graph` (`Inline` / `Left column`), and column mode behaves exactly as it does today — resizable, width remembered, double-click to reset. If you'd prefer to ship with `column` as the default instead, that's a one-word change in `DEFAULTS` and everything else in this PR still applies.

Why I think inline is worth being the default: the column reserves a fixed slice of the canvas, but most rows have no label at all — so most of the time it's empty space that the graph and commit messages could be using. Screenshots are at the bottom of this PR.

## What else is new

**You can now tell local and remote branches apart at a glance.** Label chips get a small icon: a monitor for local branches, a cloud for remotes — the same icons the sidebar already uses, drawn straight onto the canvas.

**When a local branch and its remote are on the same commit, they become one chip.** Instead of `stable` and `origin/stable` saying the same thing twice, you get a single `stable` chip with both icons — so "pushed and in sync" is visible instantly, and when the tips drift apart you see two chips on their different commits. Hovering a chip lists everything it stands for.

**Chips match the graph.** Every chip is tinted the same colour as the lane line it sits next to, in both layouts, so a label and its branch line read as one thing. The current-branch chip stays solid (everything else is a light tint), and its text automatically switches between black and white depending on the fill, so it stays readable in both themes.

**Everything you could do with labels still works inline.** Hover tooltips, right-click menus (branch menu on locals, checkout on remotes), the `+N` overflow chip with click-to-cycle — and `+N` now also shows its hidden refs on hover, which it previously didn't.

## Notes for review

- The layout switch is intentionally tiny under the hood: inline just sets the label column's width to 0, which the renderer already knew how to handle — so column mode's code path is untouched.
- The chip merging logic lives in `reforder.ts` as a plain function with its own tests (21 in that file now). The setting follows `graphLabelPriority`'s existing pattern end to end.
- Design mode includes demo rows for each chip form (merged, in-sync pair, drifted pair), so you can see all of it in a browser without a real repo.

## Testing

`svelte-check` clean · vitest 45 files / 1385 tests · `vite build` · Playwright e2e 12/12 — all green. The branch also went through several rounds of self-review before opening this PR; things caught and fixed along the way include a column-width clamp regression, a stale hover-target window after switching repos, and the light-theme contrast issue mentioned above.

Happy to split anything out or flip the default if you'd rather — thanks for taking a look! 🐱

## Screenshots
4-hover-tooltip
1-inline-default
2-settings-select
3-column-mode

### DIFF
--- a/index.html
+++ b/index.html
@@ -2153,8 +2153,9 @@ button{font-family:inherit;font-size:inherit;color:inherit;cursor:pointer}
 <!-- Tama's top-of-everything message toast — surfaces errors/results when a
      modal is covering her nook (see .top-toast CSS + main.ts's _topToast). -->
 <div class="top-toast" id="topToast" aria-hidden="true"></div>
-<!-- Full branch/tag name shown when hovering a (possibly truncated) label in the
-     graph's BRANCH/TAG column — driven from legacy/main.ts's pointermove. -->
+<!-- Full branch/tag name shown when hovering a (possibly truncated) ref label in
+     the graph — the left column's cell in column layout, a chip or its "+N"
+     pill inline — driven from legacy/main.ts's pointermove. -->
 <div class="rname-tip" id="graphLabelTip" style="display:none" aria-hidden="true"></div>
 
 <!-- ============================ DANGER MODAL (filter-repo) ============================ -->
--- a/src/islands/settings/Settings.svelte
+++ b/src/islands/settings/Settings.svelte
@@ -19,7 +19,7 @@
     TAMA_MOMENT_FIELDS,
     tamaPoseLabel,
   } from "./settings.svelte.ts";
-  import type { ThemeMode, SnapshotRetentionMode, GraphLabelPriority, TamaMotionPreset } from "./settings.svelte.ts";
+  import type { ThemeMode, SnapshotRetentionMode, GraphLabelPriority, GraphLabelLayout, TamaMotionPreset } from "./settings.svelte.ts";
   import type { ConfigScope } from "../../ipc/bindings";
   import { playTamaSound } from "../../legacy/sound.ts";
   import { updaterCtrl } from "../updater/updater.svelte.ts";
@@ -115,6 +115,19 @@
         >
           <option value="tag">Tags first</option>
           <option value="branch">Branches first</option>
+        </select>
+      </div>
+
+      <p class="mut" style="font-size:11.5px;margin:0 0 8px">
+        Inline draws a commit's ref chips right before its subject text; Left column keeps them in a separate, resizable column.
+      </p>
+      <div class="rm-form" style="margin-bottom:14px;max-width:220px">
+        <select
+          value={settingsCtrl.graphLabelLayout}
+          onchange={(e) => settingsCtrl.setGraphLabelLayout((e.target as HTMLSelectElement).value as GraphLabelLayout)}
+        >
+          <option value="inline">Inline (before the subject)</option>
+          <option value="column">Left column</option>
         </select>
       </div>
 
--- a/src/islands/settings/Settings.svelte
+++ b/src/islands/settings/Settings.svelte
@@ -106,7 +106,7 @@
       </label>
 
       <p class="mut" style="font-size:11.5px;margin:0 0 8px">
-        When a commit's labels don't all fit the gutter, show this kind first. Click a row's <b>+N</b> chip to cycle the rest into view.
+        When a commit's labels don't all fit, show this kind first. Click a row's <b>+N</b> chip to cycle the rest into view.
       </p>
       <div class="rm-form" style="margin-bottom:14px;max-width:220px">
         <select
--- a/src/islands/settings/settings.svelte.test.ts
+++ b/src/islands/settings/settings.svelte.test.ts
@@ -36,6 +36,7 @@ vi.mock("../../legacy/bridge", () => ({
   applyThemeMode: vi.fn(),
   setGraphShowAllTags: vi.fn(),
   setGraphLabelPriority: vi.fn(),
+  setGraphLabelLayout: vi.fn(),
   setTamaEnabled: vi.fn(),
   applyTamaSkin: vi.fn(),
   clearTamaSkin: vi.fn(),
@@ -243,6 +244,7 @@ describe("loadSettings / saveSettings — localStorage persistence", () => {
       soundEffectsVolume: 1,
       showAllCommitTags: false,
       graphLabelPriority: "tag",
+      graphLabelLayout: "inline",
       autoFetchEnabled: false,
       autoFetchIntervalMinutes: 15,
       autoMaintenanceEnabled: false,
@@ -270,6 +272,7 @@ describe("loadSettings / saveSettings — localStorage persistence", () => {
       soundEffectsVolume: 0.9,
       showAllCommitTags: false,
       graphLabelPriority: "tag",
+      graphLabelLayout: "inline",
       autoFetchEnabled: false,
       autoFetchIntervalMinutes: 15,
       autoMaintenanceEnabled: false,
@@ -295,6 +298,7 @@ describe("loadSettings / saveSettings — localStorage persistence", () => {
       soundEffectsVolume: 1,
       showAllCommitTags: false,
       graphLabelPriority: "tag",
+      graphLabelLayout: "inline",
       autoFetchEnabled: false,
       autoFetchIntervalMinutes: 15,
       autoMaintenanceEnabled: false,
@@ -329,6 +333,7 @@ describe("loadSettings / saveSettings — localStorage persistence", () => {
       soundEffectsVolume: 1,
       showAllCommitTags: false,
       graphLabelPriority: "tag",
+      graphLabelLayout: "inline",
       autoFetchEnabled: false,
       autoFetchIntervalMinutes: 15,
       autoMaintenanceEnabled: false,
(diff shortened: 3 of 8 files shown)


zangjiucheng (commented):
Did a thorough pass here: traced each changed function plus its callers, ran the affected unit tests (107 pass), and checked for stale references after the `orderRefs`/`paintChip`/`measureChip` signature changes. No crash-level or data issues, and the subtle parts hold up: the cross-mode `colW.graph` semantics, `mergeRefChips` pairing (including a remote listed before its local), rotation on the merged list, and the divider/hit-test changes. Nothing below blocks merging. Six lower-severity findings, most actionable first.

**1. Right-clicking the inline `+N` overflow pill opens the commit menu, not a ref menu.** Hovering that same pill shows the full ref tooltip (`labelAt` reads `overflowHit`), so it reads as a ref surface, but the contextmenu path uses `chipEntryAt`, which is null over the pill (its spans live in `overflowHit`, not `chipHit`), so `inGutter||chip` is false and you get the commit menu. It is also asymmetric with column mode, where a right-click anywhere in the gutter cell resolves a ref.
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/main.ts#L1222-L1231

**2. The inline subject can start inside the `MIN_SUBJECT_W` floor.** Chips are capped at `W - AUTHOR_GUTTER - MIN_SUBJECT_W`, then `MSG_TEXT_PAD` (11px) is added to `cx` after that cap, so a row whose chips fill to the cap leaves the subject about 29px instead of the intended 40px minimum. Cosmetic over-truncation, not a crash.
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/main.ts#L566-L575

**3. `mergeRefChips` allocates a Map and Set per labelled row on every full frame.** `displayChipsFor` runs per labelled visible row each full frame (including during fast scroll, where chips still draw), so each labelled row now pays `orderRefs` plus a fresh Map and Set plus `rotateChips`. A `length===1` fast path (or a per-row memo keyed on refSig) removes most of it on the software-rasterized canvas path.
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/reforder.ts#L67-L70

**4. `chipHit` is built in both layouts but only read inline.** In column mode `chipEntryAt` is never consulted (the contextmenu path forces `chip=null`), yet `drawGutterChips` still allocates the spans array and writes `chipHit` for every labelled row every frame. Gating it on `bcw===0` avoids the waste.
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/main.ts#L911-L913

**5. `bcol` duplicates `col`.** `bcol` is byte-identical to `col` computed earlier in the same loop iteration, so it can just reuse `col`.
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/main.ts#L564
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/main.ts#L524

**6. The composed display pipeline has no end-to-end test.** `orderRefs`, `mergeRefChips`, and `rotateChips` are each unit-tested, but `displayChipsFor`'s order (priority sort, then merge, then rotate) is never exercised as a whole, so a future reordering would pass every reforder unit test.
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/main.ts#L2006-L2009
https://github.com/zangjiucheng/GitCat/blob/ae3e9216069756bff8e3f108734c96c0e0e9999f/src/legacy/reforder.test.ts#L21-L27

Thanks for the careful work on this. Happy to see any of these land as follow-ups if you would rather merge as-is.

zangjiucheng (changes requested):

JerryHyun:
Thanks — this was a genuinely useful pass, and every one of the six is now addressed. Also merged `dev` in (the branch was 27 behind; no conflicts).

**1. Right-clicking the `+N` pill.** Fixed — the pill is a ref surface everywhere now. The span test `labelAt` had inline is a shared `overflowPillAt` helper, so the hover tooltip, the context menu and the click-to-cycle path all agree on where the pill is.

While fixing it I hit something worth flagging: routing the pill to the whole-row pool faithfully reproduced column mode's behaviour, *including* a mismatch column mode has had all along. `labelAt` builds the tooltip from the display list (sorted, merged, rotated) and bolds its first entry, but the menu resolved against raw backend order. On a commit carrying `main` (head) and `feat`, one click on the pill leaves the tooltip bolding `feat` while the menu still opened `main` — one branch highlighted, another branch's delete/reset menu. Both layouts now resolve against the display list, so the menu always names the ref the tooltip just bolded. **That does change column mode** (display-first local rather than backend-first local) — happy to split it back out if you'd rather keep this PR to the inline surface.

**2. `MIN_SUBJECT_W`.** Fixed — `MSG_TEXT_PAD` is subtracted from the chips' x-limit instead of added past it, so the 40px floor holds on every path out of `drawGutterChips` (including the `PLUS_RESERVE` re-fit and the "only draw the pill when it fits" guard). One consequence: chips get 11px less room, so `+N` appears very slightly earlier.

While checking it I found a sibling that I have *not* touched: `cx = tx + MSG_TEXT_PAD` eats the same 11px once `tx` reaches its own `W-AUTHOR_GUTTER-MIN_SUBJECT_W` clamp — on labelled and unlabelled rows alike, reachable by dragging the graph|message divider fully right. It predates the chips, and fixing it means moving that clamp, so it felt like the wrong thing to fold in here. Say the word if you'd like it in this PR.

**3. `mergeRefChips` allocations.** Took the `length===1` fast path rather than the memo — one ref is the overwhelmingly common labelled row and it can't pair with anything, so it skips the Map, the Set and the second pass. `orderRefs` got the symmetric guard too, since it was still doing decorate → sort → undecorate for that same one-ref row on the same per-frame path. Left the refSig memo alone: the fast paths take the bulk of it without a cache to invalidate.

**4. `chipHit` in both layouts.** Fixed, but as an explicit `recordHits` argument rather than a `bcw===0` test inside the function — the caller has already branched on the layout two lines up, and re-deriving it from a global would leave a future third call site silently doing the wrong thing. The `delete` stays unconditional, so a layout switch can't strand spans.

**5. `bcol`.** Gone; it reuses `col`.

**6. No test for the composed pipeline.** The composition moved into `reforder.ts` as `displayChips()`, where a test can actually reach it; `displayChipsFor` is now just the three per-row inputs. Seven new cases pin sort → merge → rotate.

One honest caveat on what that buys, because I tried to break it and mostly couldn't: swapping sort and merge is **unobservable** — both orders reduce to "the survivors of the fold, stably sorted by kind", since the fold preserves each entry's kind and first-appearance order. Rotating the raw list also coincides at small rotations, because `orderRefs` sorts remotes last, so every ref the fold removes sits in the trailing remote block and dropping them commutes with a rotation that doesn't reach past the survivors. It stops coinciding once the rotation exceeds the merged count, which one case pins. So the tests catch a wrong priority table, a broken fold, a rotation that runs before the sort, and a rotation whose modulus is the raw ref count — not every conceivable reordering, because most reorderings aren't observable. The block comment in the test file says exactly this, so nobody reads more into it later.

Two other things came out of reviewing my own fixes:

- The fast-path parity test originally asserted against copied literals, which would have happily agreed with whichever path went stale — precisely the regression it was named for. It now runs the general path and compares, with the literals kept alongside as a separate shape check.
- Two comments were claiming more than the code delivered (`drawGutterChips`' header still said it records `chipHit` "exactly like `overflowHit`"; the pipeline block claimed the folded-away refs form a *suffix*, which isn't true — an unmatched remote survives inside the trailing remote block). Both corrected downward.

`svelte-check` clean · vitest 1405 (9 new) · `vite build` · Playwright 12/12.

zangjiucheng:
LGTM; a few nips I left below, but not a blocker to merge. I want this good stuff in the v1.1 release tmr so I will directly merge it in. For the nips pls take a look if you can resolve them later. 

The `sort → merge → rotate` ordering written up as an explicit contract in `reforder.ts`, and the note in `recomputeLayout` about gating on the layout mode rather than the resulting width, are both the right calls — `colW.branch = colDrag.bcw0 + dx` really does have no floor, so that clamp has to run unconditionally.

**Merging as-is**, with inline staying the default — I want this in the next release. Four follow-ups below that I'll take myself post-merge; none of them block, and there's nothing to do on your side. Recording them here so they don't get lost.

---

**1. `dividerAt` exposes a drag zone with no repo open** — `src/legacy/main.ts:1036`

The old guard was `if (bcw <= 0) return null`, which shut off every divider. The inline branch now returns `"graph"` whenever `lastTx > COL_HANDLE`, and `lastTx` is module-level state that `closeRepo` never resets — so after closing a repo there's an invisible col-resize strip left on the canvas, and dragging it persists a `colW.graph` override. The dashboard modal covers the canvas so it isn't reachable in practice. An `if (!G) return null` closes it for good.

**2. `chipHit` allocates every frame and grows without bound** — `src/legacy/main.ts:927`

`chipHit.set(row, placed.map(c => ({x0, x1, entry})))` sits in the draw loop, so every labelled visible row allocates a fresh array plus one object per chip, per frame. The old code only wrote `overflowHit` for rows that actually overflowed, which is a much smaller set. The map is also only cleared wholesale on repo switch / cache restore, so every labelled row ever scrolled past keeps an entry for the session.

This canvas is main-thread and the dev webview software-rasterises (roughly 35× slower than GPU), so I want to profile it on a tag-heavy repo before deciding. If it shows up, the fix is to have `chipEntryAt` recompute spans on demand instead of recording them per frame. Same neighbourhood: `paintChipIcon` does a `save()`/`restore()` pair plus a transformed `Path2D` stroke per icon.

**3. Two unlabelled selects stacked under one heading** — `src/islands/settings/Settings.svelte:123`

The new layout select follows the priority select with no `<h4>` or `<label>` of its own, so "Graph" now has two bare dropdowns and two paragraphs of small text under it. Needs a sub-heading or at least a `title`. Cosmetic.

**4. `rotateChips` copies on the common path** — `src/legacy/reforder.ts:126`

With `k === 0` (rot unset, the usual case) it still does `list.slice()`, and the caller is already holding the fresh array `mergeRefChips` just built. Genuinely trivial, and it's arguably the honest thing for a pure function to do — noting it only because it's in the same per-row per-frame path as the item above.

---

Nice work on this one, and thanks for the self-review rounds before opening it — the chip-merging tests in particular made the review much faster. 🐱

zangjiucheng (approved):

PR #25, opened 2026-08-08
feat(cli): install the `gitcat` command from the app (macOS, Linux, Windows)

zangjiucheng:
## What

Makes the `gitcat` command reachable from a terminal on all three desktop platforms. The CLI feature (`gitcat <folder>`, VS Code `code .` style) already worked, but the binary lives inside the installed app and nothing put it on `PATH`. This adds a VS Code-style installer, the same way VS Code ships its own `code` command install.

- **Install 'gitcat' Command in PATH** action in ⌘K, and a **Command line** section in Settings. Both shown in-app on any OS (`IN_TAURI`).
- Backend `install_cli_shim()` (new `cli_shim.rs`) writes a small launcher to a folder on PATH.

## Per-platform install

- **macOS** — `/usr/local/bin/gitcat`. Opens the app through LaunchServices (`open -n -a`), not a direct exec of the Mach-O: `open` returns immediately so the terminal isn't blocked, and LaunchServices activates the window so its keyboard layer works from the first keystroke (a direct exec leaves it unfocused until clicked, the same trap `windows.rs` documents). Resolves a relative path against the caller's CWD first, since `open` launches from a different one. Falls back to one native admin password prompt (`osascript`) when `/usr/local/bin` isn't user-writable.
- **Linux** — `~/.local/bin/gitcat` (no root). Backgrounds the app with `nohup … &` so the terminal returns; the app inherits the shell's CWD, so relative paths resolve. Targets `$APPIMAGE` when set (its stable outer path) rather than `current_exe()`, which points into the ephemeral AppImage mount. The `.deb`/`.rpm` already put `gitcat` on PATH, so this mainly helps AppImage users.
- **Windows** — `gitcat.cmd` under `%LOCALAPPDATA%\Microsoft\WindowsApps` (user-writable, on PATH by default). Uses `start "" /D "%CD%"` so it doesn't block the terminal and relative paths resolve against the caller's directory.

Non-blocking on every platform, which was the explicit ask.

## Setup-wizard step

The installer is also offered as an optional step in the first-run wizard, VS Code onboarding style: pick → identity → **install gitcat command** → done. It reuses the same `install_cli_shim` backend, shows the installed path (or error) inline, and is fully skippable ("Not now" proceeds, Esc dismisses). Back returns to the identity step. Demo mode shows the shape of a success without a backend call.

## CI

Added a `rust-windows` job to `ci.yml`. The existing `rust` job only compiles Linux, so the `#[cfg(target_os = "windows")]` paths (this PR's `cli_shim::windows_install`, plus `windows.rs`/`procutil.rs`/`wsl.rs`) were only ever built by the tagged release matrix. The new job builds the crate on `windows-latest` and compiles every test target, running the platform-agnostic `cli_shim` launcher tests as a smoke check. Kept build-plus-smoke rather than a full `cargo test` so the PR isn't coupled to the whole suite's Windows runtime portability, which is a separate concern.

## Bonus fix: no setup wizard when opening a repo from the CLI

A `gitcat <folder>` launch (and the Dashboard's "Open in New Window") carries `?repo=` on the window URL, but `openRepo()` runs async so `bridge.CUR_REPO` is still `null` when the wizard auto-start gate runs at boot. On a machine that had not dismissed the wizard, that popped the first-run setup wizard on top of the repo the user explicitly asked to open. The gate now also checks the launch arg, so a repo-arg launch never shows the wizard.

## Testing

- The launcher-script builders are pure functions compiled and unit-tested on every platform (`cfg(any(target_os = …, test))`), so all three launchers are verified even though CI's PR job only builds Linux and only the release matrix builds Windows. `cargo test --lib cli_shim` (5) and `cargo test --lib windows` (10, covers the `pub(crate)` change) green; `export_bindings` regenerated `bindings.ts`.
- `pnpm run check` clean, `pnpm run build` clean, `pnpm test` 1377 passing.
- Manual (macOS): install is silent on a Homebrew Mac, the generated launcher passes `sh -n`, returns the terminal immediately, focuses the window, and `gitcat .` opens the current repo.

## Notes / follow-ups

- Linux `~/.local/bin` is on PATH on most distros but not guaranteed; the success path assumes it is, and the README calls this out.
- Windows relies on `%LOCALAPPDATA%\Microsoft\WindowsApps` being on PATH (default on Windows 10/11). Not verified in the PR CI job (Linux-only); it builds in the release matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,12 +168,15 @@ jobs:
   # only compiles Linux, so any `#[cfg(target_os = "windows")]` code path (e.g.
   # cli_shim::windows_install, plus windows.rs / procutil.rs / wsl.rs) is
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
-  # catches a Windows-only compile break on the PR instead. Deliberately
-  # build-plus-smoke-test rather than a full `cargo test`: `cargo test cli_shim`
-  # still COMPILES every test target on Windows (so a Windows-only test compile
-  # error is caught too) but only RUNS the platform-agnostic cli_shim launcher
-  # tests, so the PR isn't coupled to the whole suite's runtime portability on
-  # Windows (temp-path/git-behaviour differences), which is a separate concern.
+  # catches a Windows-only compile break on the PR instead.
+  #
+  # The smoke test is `--lib` (library unit tests) ONLY: those are Windows-clean
+  # (the handful of Unix-only ones are #[cfg(unix)]-guarded), and it runs the
+  # platform-agnostic cli_shim launcher tests. The integration tests under
+  # tests/ (submodule / repo_registry / repo_files) use std::os::unix APIs with
+  # no cfg guard, so they don't compile on Windows yet — making that suite
+  # Windows-portable is a separate follow-up, deliberately out of scope here so
+  # this job stays a reliable compile guard rather than a flaky full test run.
   # No apt/system-deps step: Windows runners ship the MSVC toolchain and WebView2
   # SDK, and libgit2-sys builds its vendored copy with `cc`, same as on Linux.
   rust-windows:
@@ -197,8 +200,8 @@ jobs:
       - name: Build
         run: cargo build --locked
 
-      - name: Smoke-test (compiles all tests on Windows, runs the cli_shim launcher tests)
-        run: cargo test --locked cli_shim
+      - name: Smoke-test (library unit tests, incl. the cli_shim launcher tests)
+        run: cargo test --locked --lib cli_shim
 
       - name: Build examples
         run: cargo build --examples --locked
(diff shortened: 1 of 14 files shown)


zangjiucheng:
Addressed both Cursor Bugbot findings:

**Medium — "Bundle path breaks shell script"** (`cli_shim.rs`): fixed in 6a74a7b. The launcher now single-quotes the baked-in install path (`sh_single_quote`, with the `'\''` idiom for an embedded quote), so `"`, `$`, `` ` ``, `\`, spaces, etc. in the path can no longer break the script or resolve the wrong bundle. Same treatment for the Linux `nohup` target and the osascript admin `cp`/`chmod` paths; on Windows the exe path doubles any `%` (batch's only in-quotes metacharacter — filenames can't contain the others). Added tests for a `$` in the macOS path, an embedded single quote, and a `%` in the Windows path.

**Low — "Settings shows install without Tauri"** (`Settings.svelte`): already resolved before this review landed. That comment was against the first commit (`11573cc`), where the section gated on `IS_MAC`. The cross-platform commit (`8f10442`) changed it to `{#if IN_TAURI}`, matching the ⌘K action, so neither entry point shows in browser design mode.

`cargo test --lib cli_shim` now 7 passing.

PR #28, opened 2026-08-09
Sidebar: click a branch or tag to jump to its commit in the graph

JerryHyun:
Hi! This one changes what clicking a row in the sidebar does, so let me lead with the part you'll want to weigh in on.

## The gesture change (and why I went this way)

**Clicking a branch, remote branch or tag now selects its tip commit in the graph and centres it.** Checking out moved to **double-click**, and stays on right-click and the row's ⋮ button.

Today a single click opens the checkout-confirm popover — so the most obvious gesture in the whole tree is also the one that touches your working tree. There was no way to answer "where is this branch?" by clicking it; the only route was ⌘K. Making the plain click the safe, read-only one and putting checkout behind a deliberate second gesture felt like the right trade, and it matches what most desktop git clients do. But it *is* a change existing users will notice on upgrade, so if you'd rather it landed differently — click doing both, or checkout staying on single click — say the word; the wiring is a few lines either way.

Two smaller decisions inside that:
- **The current branch's row now responds to clicks.** It was inert before (checking out the branch you're on is a no-op), but jumping to your own HEAD is useful, so the jump path has no such guard. The checkout path keeps it.
- **Remote rows gained a ⋮ button**, mirroring local and tag rows. They had neither a ⋮ nor a right-click menu, so once Enter/Space on the row meant "jump", there was no keyboard route left to check out a remote at all. The new button opens the checkout confirm directly, since checkout is a remote row's only action.

## How it finds the commit

The sidebar already knows every ref's commit id, and the graph already keeps a full-length id per row — so a click is just a lookup, with no reload and no waiting.

One detail worth naming because it's an easy trap: the row metadata carries a **7-character** short hash while a ref's id is the **full 40**, so the lookup joins on the full-id array (`BACKEND.oids`) that the incremental-refresh path already uses. Comparing against the short hash wouldn't match at all — and "fixing" that by truncating would start landing on the wrong commit once a repo gets big enough for 7 characters to collide.

A nice consequence of asking "is this commit loaded?" rather than "is this branch ticked?": an unticked branch that's already merged into a visible one is *in* the graph, so clicking it just works.

## When it can't jump

Three different situations end in "no row for this commit", and only one of them is a checkbox problem — so the message says which: the graph is still streaming; the branch is unticked (tick it); or nothing currently shown reaches that commit. That last one is the normal case for tags, which never seed the graph walk and have no checkbox to tick.

## A fix that came along for the ride

Writing the end-to-end test turned up something pre-existing: **the Playwright mock never actually delivered graph data.** It answered `load_graph`'s call, but the real payload arrives over a `graph-batch` event — so every mock-backed test had been running against an empty graph without anyone noticing. The mock now fires that event with the real batch shape, which is what makes it possible to test anything graph-related at all.

## Testing

- 8 unit tests on the lookup and each failure message (including that a tag is never told to tick a checkbox).
- 5 end-to-end cases over a real temporary repo: click jumps; double-click still confirms checkout; Enter on the ⋮ opens the menu rather than jumping; Enter on the row itself still jumps; a remote's ⋮ opens the confirm.
- `svelte-check` clean · vitest 1384 · `vite build` · Playwright 17/17.
- Also driven by hand in the real app on Windows — worth doing, since the e2e runs in Chromium and this touches focus behaviour that WebView2 could have differed on.

Happy to adjust the gesture split or anything else — thanks for taking a look! 🐱

## Demo

https://github.com/user-attachments/assets/2f942d1e-8f97-4db3-836a-71447357cccf


JerryHyun:
@zangjiucheng 
### Please sanity-check my Chinese 🙏

Merging `dev` brought the i18n work onto this branch, and it turned out `sidebar.svelte.ts` merged with **no conflict at all** — which was the trap. You'd keyed every Tama toast in that file; this branch had added new ones afterwards, so git had no reason to touch them and they'd have shipped as English holes in the Chinese UI. Same for a remote row's ⋮ button, which is new here.

So there are six new keys, plus one existing string whose English source changed. I wrote the `zh/` side myself and **I don't read Chinese** — the English is the part I can stand behind. Please rewrite anything that reads wrong; I won't be offended, and I'd rather you change it than leave it.

| key | `en/` (source of truth) | `zh/` (mine — please check) |
|---|---|---|
| `sidebar.jump_no_commit` | `{name} has no commit to jump to.` | `{name} 没有可跳转的 commit。` |
| `sidebar.jump_still_loading` | `Still loading the graph — try {name} again in a moment.` | `图还在加载中 —— 请稍后再点一次 {name}。` |
| `sidebar.jump_not_shown` | `{name} isn't shown in the graph — tick its checkbox to load it.` | `{name} 没有显示在图中 —— 勾选它的复选框即可加载。` |
| `sidebar.jump_not_reachable` | `{name} isn't in the loaded graph — no branch currently shown reaches its commit.` | `{name} 不在已加载的图中 —— 当前显示的分支都到不了它的 commit。` |
| `sidebar.checkout_remote` | `Checkout this remote branch` | `检出这个远程分支` |
| `sidebar.checkout_named` | `Checkout {name}` | `检出 {name}` |

Two notes on the wording, in case they affect how you'd translate:

- The first four are the failure messages for clicking a ref that the graph can't jump to. Three different situations all end in "there's no row for that commit" and **only one of them is a checkbox problem**, so each says which — that distinction is the whole point of having four strings instead of one. `jump_not_reachable` in particular is the *normal* case for a tag: tags never seed the graph walk and have no checkbox to tick, so telling a user to "tick its checkbox" would be advice they can't follow.
- `checkout_remote` / `checkout_named` are a verb where a local branch's button just says "Branch actions", because checking out is a remote row's only action — its ⋮ opens the confirm directly rather than a menu.

**`help.branches_p` also needed updating in both locales**, not just merging: this PR changes what a click does, so the old sentence ("click one to check it out") became wrong in every language. The English now reads "click one to jump the graph to its tip commit; for a branch, double-click (or right-click, or its ⋮ button) checks it out instead" — my zh rendering is `点击即可让图跳转到它的顶端 commit;对分支而言,双击(或右键点击、或它的 ⋮ 按钮)则改为检出`. Worth a second look, since it's the longest of the lot.

zangjiucheng:
Lovely feature, I will take a look today :D  BTW, translation is not priority task for now (But good to have; even just en support should also be good enough). Focus on feature dev still. I just invited t() to translate the feature roadmap.

zangjiucheng:
Nice work. On the gesture question you led with: **your way, as built.** Click = navigate, double-click = checkout is what Fork/Sourcetree/GitKraken all do, and it fits GitCat's safety-first line better than the status quo — the most inviting gesture in the tree shouldn't be the one that touches the working tree.

### Few change request before merge in:

**1. Design mode can never jump, so every ref click warns** — `src/legacy/main.ts:2509`

```js
if(BACKEND){ G=buildGFromBackend(BACKEND); … }
else { G=generateGraph(N); … }        // design mode: BACKEND stays null
```

```js
function goToOid(oid){
  if(!oid||!BACKEND||!BACKEND.oids) return false;   // ← always false in design mode
```

And independently, the demo refs couldn't match anyway (`sidebar.svelte.ts:145`):

```ts
{ name: "main", sha: "a1b2c3d", … }   // 7 chars, and disjoint from generateGraph's own labels
```

Confirmed in a browser — clicking the demo `main` row gives *"main isn't in the loaded graph — no branch currently shown reaches its commit."* Before this PR that click opened the checkout-confirm popover, which demoed fine. You had design mode in mind for `graphStreamComplete`; this is the other half of that thought.

**2. A ticked branch that isn't in the graph gets a self-contradicting message** — `sidebar.svelte.ts:835`

```ts
if ((section === "local" || section === "remote") && !this.isBranchVisible(section, name)) {
  bridge.tama.warn(t("sidebar.jump_not_shown", { name }));
  return;
}
bridge.tama.warn(t("sidebar.jump_not_reachable", { name }));
// "{name} isn't in the loaded graph — no branch currently shown reaches its commit."
```

Execution only reaches that last line after `isBranchVisible` passed, so the branch *is* shown and the sentence argues with itself. Two real states get there, both setting `graphStreamComplete = true` over an incomplete graph: a truncated load (`lastLoadTruncated`, `main.ts:2560`) and a revwalk that errored partway (`main.ts:2776-2783`). Fix: export `lastLoadTruncated` too and add a fourth message.

**3. Double-click runs the jump twice before opening the confirm** — `Sidebar.svelte:324`

```svelte
onclick={…  sidebarCtrl.jumpToRef("local", b.name, b.sha); }
ondblclick={… sidebarCtrl.openCheckoutConfirm(b.name, false, r.left, r.bottom + 4); }
```

The DOM fires click, click, dblclick — so a checkout-by-double-click also selects the tip twice (two `commit_detail` round-trips via `detailCtrl.select`) and moves focus to the canvas first. Mostly harmless; the awkward case is double-clicking an *unticked* branch, where you're told to tick a checkbox before being offered the switch.

**4. A tag row's Enter no longer opens the tag menu** — `Sidebar.svelte:628`

```svelte
- onkeydown={(e) => (e.key === "Enter" || e.key === " ") && !sidebarCtrl.busy && sidebarCtrl.openTagMenu(…)}
+ onkeydown={(e) => { … sidebarCtrl.jumpToRef("tag", tag.name, tag.sha); }}
```

Two description nits: the `jumpToRef` block has 7 tests, not 8; and the graph-batch mock fix landed on `dev` separately (2273853 here is "take its graph-batch mock, drop my duplicate"), so it isn't in this diff.

On the Chinese — it's good, ship it. Matches the conventions the zh locale already set: `图` for the graph, `检出`, `复选框`, `commit` left untranslated. I wouldn't change a word.

Nice work, especially the keyboard half — `e.target !== e.currentTarget` with e2e tests pinning *both* directions of it is the kind of thing that usually only gets written after someone breaks it. 🐱

zangjiucheng (changes requested):
See note above

JerryHyun:
Thank you for such a careful read — you didn't just point at the four problems, you showed me where each one comes from, which made them quick to fix. All four are done, plus the two description corrections. Pushed as `5627fa3`.

## 1. Design mode could never jump

You were right, and it was worse than one bug: the jump needs commit ids, design mode has none, and the demo refs' made-up shas wouldn't have matched anything anyway. Before this PR that click opened the checkout popover and demoed nicely, so I'd made the demo worse. Sorry about that.

Design mode has no commit ids, but it does have ref names, so there's now a small `goToRefLabel(name)` next to `goToOid` that looks a row up by name instead. It only runs in design mode — with a real repo, matching by commit id is the trustworthy answer, and falling back to names there could quietly land you on the wrong row. I put that reason in a comment so it doesn't look like an arbitrary check later.

**One thing to know: this fixes some demo rows and not others.** The sidebar's demo refs and the graph's demo labels are two separate made-up lists that were never written to match each other. So in design mode now:

- **jumps:** `main`, `feat/inline-diff`, `fix/lane-cull`, `origin/main`
- **still warns:** the demo tags, `release/*`, and the `upstream/*` remotes — nothing in the generated graph is called those

Making the rest work would mean editing the demo data so both lists agree, which felt like a separate change rather than something to slip into this one. Happy to do it if you'd like. Either way I've noted the overlap in the code so nobody files the leftovers as a new bug.

There's a new `e2e/sidebar-jump-design.spec.ts` covering this in the browser, with no Tauri mock.

## 2. The message that argued with itself

Also right, and your reading of the two states was exactly it. There's a fourth message now, shown before the "not reachable" one:

> Only part of the history is loaded — {name}'s commit may be older than what was loaded.

It's driven by a new `graphIncomplete` flag, set when a load is either truncated or ends with an error. I kept it separate from `lastLoadTruncated` because that one means something narrower and two other checks already rely on it.

**One note on the approach you suggested.** I tried clearing the flag when a graph is restored from cache, but it turned out not to work: the cache accepts a load whose history walk stopped early with an error, so clearing on restore would treat an incomplete graph as complete. I ended up storing the flag in the cache entry instead, which keeps it correct either way.

While I was in there I noticed something similar just next door, which I have **not** touched: `tryFastRefresh` also reads `lastLoadTruncated`, so a load that ended in an error still keeps the quick-refresh path switched on over a partly-loaded history. It looks like the same situation, one check over. Changing it would affect reload behaviour well beyond this PR, so I've only left a `KNOWN GAP` note there — entirely up to you whether it's worth following up.

## 3. Double-click

This one turned into three separate things, and they didn't all end the same way, so here's what I found rather than just what I changed.

**The double jump — fixed.** The click handler now ignores the second click of a double-click, on all three row types. The test counts the `commit_detail` calls: two before, one after.

**The focus move — I looked into it, and it settles by itself.** Once a double-click finishes, focus is on the sidebar row, not the canvas. The first click does move it to the canvas, but the second click's mousedown brings it straight back. So there was nothing to change here — I'd rather show you that than claim credit for it.

**The scolding that arrived just before the "Switch to …?" box — fixed for normal double-click speeds.** Double-clicking a branch you're not on opens that confirm box. But the first click's jump had already failed and told you to go tick a checkbox, so a beat later the box turned up offering exactly the switch you'd just asked for. A failed jump's warning now waits a moment before appearing, and any double-click or right-click cancels it. A successful jump is still instant, and there's a test making sure success never sets a timer at all.

I first used 250ms, which was too short — the browser decides what counts as a double-click using the system's own setting, which is usually 500ms. Measuring by the gap between the first click and the double-click:

| gap | warning |
|---|---|
| 1ms | doesn't appear |
| 408ms | doesn't appear |
| 816ms | appears |
| 1436ms | appears |

So on a normally configured machine it works, and past 500ms it doesn't. Someone who has slowed their double-click speed in accessibility settings would still see the old sequence. Unfortunately no test can catch that — Playwright's `dblclick()` has no gap between its clicks, so the failing range isn't reachable from a spec. The boundary and the reason are both in a comment.

**Two trade-offs I'd rather mention than let you discover:**

- A failed jump's warning is now half a second late, including on plain single clicks that were never going to be double-clicked.
- The 500ms is a guess, because a web page can't read the system's double-click setting. Since this is Tauri, that value *is* available natively on all three platforms, so if you'd ever prefer it exact instead of approximately right, a small command would do it. Very much your call, and not something for this PR.

## 4. Enter on a tag row

I've left this one as jumping — but it's your design, so if you'd rather it opened the menu, just say and I'll change it.

My thinking: Enter matching what a click does is the point of the split you approved, and the tag menu is still there via the row's ⋮ button and right-click, the same as a branch row. What was genuinely missing was proof — the tests covered a *branch* row's ⋮ but not a tag row's. That case exists now, and checks both that the menu opens and that the graph stays put.

## Something small I noticed nearby

The checkout popover doesn't take keyboard focus when it opens, while the new-branch, new-tag and push popovers all do. So with it open, focus stays on the sidebar row and Enter re-runs the row's action instead of confirming; Switch is only reachable by Tab.

It isn't something this PR introduced and it wasn't on your list, so I've left it alone — just flagging it since it sits right next to code I was already in.

## Testing

`svelte-check` clean · vitest 1421 · `vite build` · Playwright 21/21.

Two notes about running the suite, both of which I found already present on `dev` — mentioning them only so a red `pnpm test` doesn't look like something this branch did:

- `pnpm test` passes with `--no-file-parallelism` and fails without it. In the default parallel mode, `sound.test.ts` and `plumbing.svelte.test.ts` fail; both check that they haven't pulled in `legacy/main.ts`, both pass on their own, and both do the same thing on a clean `dev` checkout.
- Playwright needs `--workers=1` on my machine to be reliable; at that setting it's 21/21.

zangjiucheng (approved):
All four points from the last round are properly closed, and a couple of them went further than I asked. `goToRefLabel` with the deliberate `BACKEND` guard is the right shape for the design-mode gap — and covering it with a real-browser spec rather than a unit test is the part I'd have skipped. Carrying `graphIncomplete` through the graph cache instead of hardcoding `false` on restore is the kind of detail that usually gets found six months later.

I verified the merge against current `dev` too, not just the head. PR #47 deleted `chipHit` out from under `restoreGraphFromCache`, which is exactly where this branch adds a line — the 3-way merge resolves it correctly, `index.html` survives intact, and the merged tree is green: 1423 tests / 45 files, svelte-check 534 files 0/0, and all 8 sidebar-jump e2e specs pass.

Approving. One non-blocking nit below — take it here or leave it to me.

---

**A held warning survives a later successful jump.** The cancel lives inside `warnJumpFailed`, so the two success paths return without clearing a warning an earlier click already armed:

https://github.com/zangjiucheng/GitCat/blob/5627fa3058b3a53dd63a6c7c183e0ef8af9d3d68/src/islands/sidebar/sidebar.svelte.ts#L833-L846

Click a ref that isn't loaded, then click a different one within the hold window and land on it — 500 ms later Tama scolds you about the first ref while you're looking at the second:

```
sidebarCtrl.jumpToRef("tag", "v0.9.0", …);   // not loaded -> warning armed
vi.advanceTimersByTime(100);
sidebarCtrl.jumpToRef("tag", "v1.0.0", …);   // jumps fine
vi.advanceTimersByTime(2000);
// -> "v0.9.0 isn't in the loaded graph — no branch currently shown reaches its commit."
```

Same class of stale-scolding the hold was added to prevent, just across two refs instead of inside one double-click. Fix is to move the cancel from `warnJumpFailed` up to the top of `jumpToRef`, so every outcome — success included — supersedes whatever was pending.

Nice work on this one. 🐱

PR #40, opened 2026-08-11
Translate the pinned Uncommitted-changes band (English in every locale today)

JerryHyun:
# The pinned "Uncommitted changes" band was never translated

> **Base branch:** `dev` · Branch: `fix/canvas-band-i18n`

Found while starting work on a Korean locale, but it isn't a Korean problem — **中文 has it today.**

## The bug

The band pinned across the top of the graph draws hardcoded English:

```js
ctx.fillText("Uncommitted changes", x+dotR+14, y);
const badge = nConflict ? (nConflict+" conflicted")
            : (nStaged||nUnstaged) ? (nStaged+" staged · "+nUnstaged+" unstaged")
            : "clean";
```

The keys already exist, and `Workdir.svelte` already uses them — `workdir.uncommitted_changes`, `n_conflicted`, `staged_unstaged`, `clean`. Only this canvas copy bypasses `t()`. So the most prominent row in the app stays English in every locale, right above a fully translated graph.

Four call sites, no new strings, same branch order. One small consequence worth naming: the `·` separator moves out of a JS literal into the locale strings, so it becomes a translator's choice — `zh` keeps it.

Nothing else was needed to make a live language switch work. `drawWorkdirBand` runs from `present()`, outside the blit buffer, so the scroll fast-path can't leave stale English pixels behind, and the `dirty=true` your `i18nEvents` handler already sets brings it straight back.

## A test for canvas text

The band is pixels — no node, no accessible name. That's precisely how it stayed English through every locale you've shipped without anyone noticing, so it seemed worth leaving something behind that would notice next time.

`e2e/band-i18n.spec.ts` hashes the band's row out of the canvas backing store and asserts it changes with the language, switching through the Settings picker rather than a test hook. That also exercises the live-switch path your `i18nEvents` handler promises in its own comment — as far as I could tell, nothing was covering it.

A few decisions in there that aren't obvious:

- **It reads `getImageData`, not a screenshot.** A screenshot physically cannot see the badge: the dev perf HUD is positioned over the canvas's top-right corner, exactly where the badge is right-aligned, and rewrites its fps readout several times a second.
- **The row is hashed in two halves and both are asserted.** The label is on the left, the badge on the right; one whole-row hash is satisfied by either one changing. I only found that out by reverting the badge alone and watching a single-hash version pass.
- **It never asserts a specific hash value.** A golden number would break on any font, theme or antialiasing change, and everyone would learn to delete the test.

Two things bit me while writing it, both of which made a broken build pass green, so they're commented in place:

- `.scrim` delays its visibility flip by 160ms (`transition: … visibility 0s linear .16s`). Waiting on the class rather than on `toBeHidden()` silently loses the click that follows — the band just never gets selected.
- The badge repaint is asynchronous. Capturing EN with the old badge and ZH with the new one makes both halves differ whether or not anything is translated.

What it still doesn't reach is `workdir.n_conflicted` — the demo status is fixed at `conflicted: 0`, so that branch needs a real repo in a conflicted state. That's stated in the file rather than left for someone to discover.

## Second commit: the canvas couldn't see `--mono`

Separate commit, and separable — the translation doesn't depend on it. Three canvas sites hardcoded `"…px ui-monospace,monospace"` while the DOM reads `var(--mono)`, so the graph and the diff view could disagree about what "monospace" means. They now share one constant read from the same variable.

**This one has a visible effect and I'd rather flag it than have you find it.** Wherever the named faces are actually installed, the sha column moves off the generic monospace face onto the first of them and gets slightly wider — that's every stock Windows box (Windows Terminal ships Cascadia Code) and any Linux box carrying `cascadia-code` or the MS core fonts. macOS already resolved `ui-monospace` to SF Mono, so it's unchanged. `AUTHOR_GUTTER` reserves comfortably more than the wider sha, so nothing collides, but it is a restyle rather than a no-op. **Happy to drop this commit if you'd rather not take it** — the first one stands alone.

I originally justified it as CJK coverage. That was wrong and I want to correct it rather than let it sit in a commit message: `--mono` names no CJK family, and Blink's per-character system fallback draws those either way — 个 measures the same under the old stack, the new one, and a deliberately nonexistent family. Naming the stack decides *which* face draws non-Latin text, not whether it draws.

## Two things I noticed but did not touch

- **`FONT_MONO` is read once at module load.** Fine today, since `--mono` is one unscoped declaration — but it sits outside `readTheme()`'s re-read path, so scoping `--mono` per theme or per locale would need a re-read there. The comment says so. It's also the only thing standing between this codebase and a user-selectable font, if that's ever of interest: `main.ts`'s locale handler already forces the repaint, so it's the read that would need to move, not the plumbing.
- **`src/legacy/sound.test.ts` looks flaky**, independent of this branch. On a pristine `dev` it failed once in three full-suite runs — same two tests, the first timing out inside `await import("./sound.ts")` rather than on an assertion. Not touched here; mentioning it in case it's been showing up in CI.

## Testing

`svelte-check` clean (534 files) · vitest 1405 · `vite build` · Playwright 13/13.

Each half of the new test was verified by injecting its regression separately: re-hardcoding the label fails on `LABEL`, re-hardcoding `staged_unstaged` fails on `BADGE`.

### DIFF
--- a/e2e/band-i18n.spec.ts
+++ b/e2e/band-i18n.spec.ts
@@ -1,0 +1,156 @@
+import { test, expect, type Page } from "@playwright/test";
+
+// The pinned "Uncommitted changes" band is CANVAS text — no node, no accessible
+// name — which is how it stayed hardcoded English through every locale the app
+// shipped without anyone noticing. It is still reachable: hash the pixels of the
+// row it occupies and check they change with the language.
+//
+// Read from the backing store rather than page.screenshot, because a screenshot
+// physically cannot see the badge: the dev perf HUD is absolutely positioned
+// over the canvas's top-right corner, exactly where the badge is right-aligned,
+// and rewrites its fps readout several times a second. getImageData sees the
+// canvas alone.
+//
+// The assertion is "the hash changes", never "the hash is <value>". A golden
+// number here would break on any font, theme or antialiasing change and teach
+// everyone to delete it.
+//
+// The row is hashed in two halves and BOTH are asserted, because the band draws
+// two independent strings: the label on the left, the status badge right-
+// aligned at W-14. One whole-row hash is satisfied by either one changing, which
+// is not the guarantee this test is for.
+//
+// That split has less headroom than it looks. By the time of the comparison the
+// badge reads "1 staged · 2 unstaged" rather than "clean", so its ink starts
+// around x375 against a midpoint near x313 — roughly 60px, not the ~250 the
+// short badge would suggest. A verbose translation of staged_unstaged is what
+// erodes it first (the English string alone spans ~236px); a much narrower
+// canvas would do it too. Either way the badge crosses into the left half and
+// the two-half split quietly degenerates into the whole-row hash it replaced.
+//
+// WHAT THIS DOES NOT COVER: workdir.n_conflicted — the demo status is fixed at
+// conflicted:0, so that branch needs a real repo in a conflicted state. And
+// workdir.clean is drawn once, in English, before the band is selected below —
+// never in zh, and never inside an assertion.
+
+// Source of truth is ROW_H_BASE in src/legacy/main.ts; the band is exactly one
+// row at zoom 1. Nothing ties these together, so a change there silently makes
+// this hash the wrong strip and click at the wrong y.
+const BAND_CSS_H = 26;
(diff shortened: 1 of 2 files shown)


zangjiucheng (approved):
LGTM, taking both commits.

PR #41, opened 2026-08-11
fix: rotateChips copies the list on the common (rot=0) path

### LINKED ISSUE #32: rotateChips copies the list on the common (rot=0) path
Follow-up from #18 (inline ref labels). Micro-cleanup, non-blocking. Line numbers are against PR head `a356f73`.

`src/legacy/reforder.ts:126`

```ts
export function rotateChips<T>(list: readonly T[], rot: number): T[] {
  const n = list.length;
  if (n === 0) return [];
  const k = ((rot % n) + n) % n;
  return k === 0 ? list.slice() : list.slice(k).concat(list.slice(0, k));
}
```

`k === 0` — rotation unset, which is every row the user hasn't clicked a `+N` pill on, i.e. essentially all of them — still copies. The only caller is `displayChips`:

```ts
return rotateChips(mergeRefChips(orderRefs(refs, tagsFirst)), rot);
```

and `mergeRefChips` already returns a freshly built array, so the copy protects nothing. That's one throwaway array per labelled visible row per frame.

Genuinely trivial, and "a pure function always returns a copy" is a defensible contract to keep — the alternative is documenting that `displayChips` may return the array `mergeRefChips` built, which is true but easy to forget. Filing it because it sits in the same per-row per-frame path as the `chipHit` allocation; if that one measures as noise, this one certainly does, and both can be closed together.

webbrain-one:
Closes #32

Patch generated by `qwen3.6-35b-a3b via local API`.

### DIFF
--- a/src/legacy/reforder.ts
+++ b/src/legacy/reforder.ts
@@ -123,7 +123,7 @@ export function rotateChips<T>(list: readonly T[], rot: number): T[] {
   const n = list.length;
   if (n === 0) return [];
   const k = ((rot % n) + n) % n;
-  return k === 0 ? list.slice() : list.slice(k).concat(list.slice(0, k));
+  return k === 0 ? list : list.slice(k).concat(list.slice(0, k));
 }
 
 // The whole display pipeline in one place: priority sort, THEN fold local +


zangjiucheng (changes requested):
Thanks for your participation in GitCat!! Pls, resolve the type error in the CI failure.

PR #48, opened 2026-08-14
feat(graph): double-click the workdir row to open its expanded diff (closes #46)

### LINKED ISSUE #46: Uncommitted changes row: double-click should open the full-page diff popup like commits
## Summary

Double-clicking the pinned **Uncommitted changes** row in the graph should open the full-page diff popup, the same way double-clicking a regular commit row already does. Right now nothing happens on the workdir row.

## Current behavior

- Double-click a **commit** row → the full-page (near-fullscreen) diff popup opens via `detailCtrl.expandDiff()`.
- Double-click the pinned **Uncommitted changes** row → nothing happens.

The workdir panel *already has* the same expanded-diff popup. You can open it today by selecting the Uncommitted changes row and clicking the expand button in the panel header. It just isn't reachable with the double-click gesture that commits use, so the two rows behave inconsistently.

## Expected behavior

Double-clicking the Uncommitted changes row selects the workdir and opens its expanded diff popup, matching how commit rows behave.

## Where the code is

- `src/legacy/main.ts` ~1261 — the canvas `dblclick` handler. It bails early on `if (!hit || hit.row < 0) return;`, which excludes the workdir band (its `hitTest` sentinel is `row === -2`, see main.ts:1031). Commit rows go on to `select(hit.row); detailCtrl.expandDiff();`.
- `src/legacy/main.ts:1803` — `selectWorkdir()` already selects the workdir panel (this is what a single click on the band calls at main.ts:1227).
- `src/islands/detail/detail.svelte.ts:520` — `expandDiff()` on the detail controller, the pattern to mirror.
- `src/islands/workdir/Workdir.svelte` ~95 — the expanded-diff modal (`diffExpanded` state) and the expand button (~277) that opens it today. This state is currently local to the component and is not exposed on the controller.

## Suggested approach

1. Lift the popup toggle onto `workdirCtrl` (in `src/islands/workdir/workdir.svelte.ts`) as an `expandDiff()` method backed by a reactive field, mirroring `detailCtrl.expandDiff()` / `diffExpanded`.
2. Have `Workdir.svelte` read that controller field instead of (or in addition to) its local `diffExpanded`.
3. In the `dblclick` handler in `main.ts`, handle `hit.row === -2`: call `selectWorkdir()` then `workdirCtrl.expandDiff()`, before the existing `hit.row < 0` guard returns.

## Why this is a good first issue

Small and well scoped, no backend or Rust changes, one gesture wired to an already-existing popup. It's a clean way to learn the canvas hit-testing flow (`hitTest` sentinels, the `dblclick`/`pointerup` handlers) and the island controller pattern (`detailCtrl` vs `workdirCtrl`). Verify with `pnpm run check` and by double-clicking the Uncommitted changes row on a dirty repo.

dchaudhari7177:
Closes #46.

Double-clicking a commit row opened the full-page diff popup; the pinned **Uncommitted changes** row did nothing. The popup already existed for the workdir — reachable from the expand button in the panel header — it just wasn't on the double-click gesture.

Followed the suggested approach exactly:

**1. `workdir.svelte.ts`** — `diffExpanded` plus `expandDiff()` / `collapseDiff()` on `WorkdirState`, mirroring `detailCtrl`'s pair.

**2. `Workdir.svelte`** — reads the controller field *instead of* its own local `$state`, not in addition to it. Two sources of truth for one popup is how it ends up half-open. The `Escape` handler and the header expand/close buttons now go through the controller too.

**3. `main.ts`** — the `dblclick` handler handles the workdir sentinel **before** the `row<0` guard:

```js
const hit=hitTest(p.x,p.y);
if(!hit) return;
if(hit.row===-2){ selectWorkdir(); workdirCtrl.expandDiff(); return; }
if(hit.row<0) return;
select(hit.row);
detailCtrl.expandDiff();
```

Same select-then-expand order the commit path uses. `workdirCtrl` was already imported there (`main.ts:9`), so no new coupling.

Lifting the state onto the controller is the part that makes this possible at all — `main.ts` has no handle on a component's local `$state`.

## Tests

5 added to `workdir.svelte.test.ts`: default closed, the open/close round trip, `expandDiff` idempotent, `collapseDiff` on an already-closed popup, and that the methods live on the controller rather than only inside the component (which is the property the canvas depends on).

## Gate

- `pnpm test` — **1407 → 1412 passed** across 45 files
- `pnpm check` (svelte-check) — 534 files, **0 errors, 0 warnings**

## What I could not verify

The gesture itself in a running app — that needs a Tauri build. The controller half is under test; the canvas branch I've only verified by reading, so it's worth a click-through on your side before merging.

### DIFF
--- a/src/islands/workdir/Workdir.svelte
+++ b/src/islands/workdir/Workdir.svelte
@@ -92,10 +92,12 @@
   // commit view uses (see Detail.svelte's .diffx) — a bigger window for reading
   // a large working-tree change, with the staged/unstaged file lists on the left
   // and the full per-hunk / per-line staging kept intact on the right.
-  let diffExpanded = $state(false);
+  // State lives on the controller so the canvas can open this too: double-
+  // clicking the pinned workdir band calls workdirCtrl.expandDiff(), the same
+  // way double-clicking a commit row calls detailCtrl.expandDiff().
   function onKeydown(e: KeyboardEvent) {
-    if (e.key === "Escape" && diffExpanded) {
-      diffExpanded = false;
+    if (e.key === "Escape" && workdirCtrl.diffExpanded) {
+      workdirCtrl.collapseDiff();
       e.stopPropagation();
     }
   }
@@ -274,7 +276,7 @@
       <div class="wd-sec-head">
         <h4 class="d-lab" style="margin:0">{t("workdir.diff")}</h4>
         {@render workdirLinesBar(file)}
-        <button class="wd-act" title={t("workdir.expand_diff")} aria-label={t("workdir.expand_diff_aria")} onclick={() => (diffExpanded = true)}>
+        <button class="wd-act" title={t("workdir.expand_diff")} aria-label={t("workdir.expand_diff_aria")} onclick={() => workdirCtrl.expandDiff()}>
           <Maximize2 class="ico" size={13} aria-hidden="true" />
         </button>
       </div>
@@ -356,7 +358,7 @@
        the right. A top-level scrim so it's a direct child of .detail, same as
        the commit modal (index.html's `.detail.collapsed>*:not(.scrim)` exempts
        it from the Focus-mode panel collapse). -->
-  <div class="scrim" class:on={diffExpanded}>
+  <div class="scrim" class:on={workdirCtrl.diffExpanded}>
     <div class="modal diffx">
       <div class="modal-head">
         <div class="diffx-head-main">
@@ -431,7 +433,7 @@
         </div>
       </div>
       <div class="modal-foot">
-        <button class="btn ghost" onclick={() => (diffExpanded = false)}>{t("common.close")}</button>
+        <button class="btn ghost" onclick={() => workdirCtrl.collapseDiff()}>{t("common.close")}</button>
       </div>
     </div>
   </div>
--- a/src/islands/workdir/workdir.svelte.test.ts
+++ b/src/islands/workdir/workdir.svelte.test.ts
@@ -1316,3 +1316,66 @@ describe("folder collapse state (Collapse all / Expand all)", () => {
     expect(workdirCtrl.isDirCollapsed("unstaged", "src")).toBe(false); // same path, different section
   });
 });
+
+describe("expanded diff popup", () => {
+  beforeEach(() => {
+    workdirCtrl.collapseDiff();
+  });
+
+  it("starts closed", () => {
+    expect(workdirCtrl.diffExpanded).toBe(false);
+  });
+
+  it("expandDiff opens it and collapseDiff closes it", () => {
+    workdirCtrl.expandDiff();
+    expect(workdirCtrl.diffExpanded).toBe(true);
+    workdirCtrl.collapseDiff();
+    expect(workdirCtrl.diffExpanded).toBe(false);
+  });
+
+  it("expandDiff is idempotent", () => {
+    workdirCtrl.expandDiff();
+    workdirCtrl.expandDiff();
+    expect(workdirCtrl.diffExpanded).toBe(true);
+  });
+
+  it("collapseDiff on an already-closed popup is a no-op", () => {
+    workdirCtrl.collapseDiff();
+    expect(workdirCtrl.diffExpanded).toBe(false);
+  });
+
+  it("lives on the controller so the canvas can open it without the component", () => {
+    // The point of lifting it out of Workdir.svelte: main.ts's dblclick
+    // handler reaches this without the panel being mounted.
+    expect(typeof workdirCtrl.expandDiff).toBe("function");
+    expect(typeof workdirCtrl.collapseDiff).toBe("function");
+  });
+
+  // Living on the controller costs what a component-local `$state(false)`
+  // gave for free: Detail.svelte destroys the whole <Workdir /> instance on
+  // deselect, so the old flag reset on every remount. A controller field
+  // survives, and would re-open the modal by itself on the next single
+  // click — on the empty "select a file" placeholder, no less.
+  it("deselect() closes the popup so it does not spring back open", () => {
+    workdirCtrl.selected = true;
+    workdirCtrl.expandDiff();
+    workdirCtrl.deselect();
+    expect(workdirCtrl.diffExpanded).toBe(false);
+  });
+
+  it("select() closes the popup left open by a previous selection", () => {
+    workdirCtrl.expandDiff();
+    workdirCtrl.select("/repo");
+    expect(workdirCtrl.diffExpanded).toBe(false);
+  });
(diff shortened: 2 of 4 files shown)


zangjiucheng (changes requested):
Approach is exactly right, and the gate checks out on my side too.

One thing to fix before this goes in, then two notes.

1. The popup reopens by itself

detailCtrl clears diffExpanded on every lifecycle exit — this is the one that matters:

https://github.com/zangjiucheng/GitCat/blob/2b496cb74b02c692ca75dc5262fa959ca8debd16/src/islands/detail/detail.svelte.ts#L601-L605

The new WorkdirState never clears it outside collapseDiff():

https://github.com/zangjiucheng/GitCat/blob/2b496cb74b02c692ca75dc5262fa959ca8debd16/src/islands/workdir/workdir.svelte.ts#L376-L393

That's load-bearing, because Detail.svelte destroys the whole <Workdir /> instance on deselect:

https://github.com/zangjiucheng/GitCat/blob/2b496cb74b02c692ca75dc5262fa959ca8debd16/src/islands/detail/Detail.svelte#L116-L117

So the old component-local $state(false) reset on every remount; a controller field doesn't. And main.ts:1835's select(row) calls workdirCtrl.deselect().

Repro: double-click the band → popup opens → click any commit row → popup gone → single-click the band → the fullscreen modal springs open on its own, on the empty "select a file" placeholder.

Fix is one line each, in select() and deselect():

this.diffExpanded = false;

2. Why every line of workdir.svelte.ts shows as changed

Not your doing, but it needs a re-commit. The file went in with CRLF on all 1256 lines; the real diff is +14:

git diff --stat origin/dev...HEAD    → 1256 insertions, 1242 deletions
git diff --stat --ignore-cr-at-eol   →   14 insertions, 0 deletions

.gitattributes should have caught that:

https://github.com/zangjiucheng/GitCat/blob/2b496cb74b02c692ca75dc5262fa959ca8debd16/.gitattributes#L8

but text=auto makes git sniff text-vs-binary, and a single NUL byte anywhere in the file flips it to binary — at which point eol=lf is skipped entirely. This file has 5 literal NULs used as key separators (L263, L292, L555), so it's exempt:

$ git ls-files --eol src/islands/workdir/workdir.svelte.ts
i/-text  w/-text  attr/text=auto eol=lf

That's why only this one of your four files flipped. Could you re-commit it with LF? Otherwise git blame for the whole file lands on this commit. src/main.ts and remotes.svelte.ts have the same hole — the .gitattributes side is mine to fix, I'll take it.

3. Two notes, nothing to do

- The gesture always opens on an empty pane — select() clears selectedDiffFile and nothing re-picks one, so you land on "select a file" instead of a diff the way the commit row does. Follow-up on me.
- hitTest() only ever returns null, -2, or 0..G.N-1, so if(hit.row<0) return; is now unreachable. Harmless, but the comment above it cites "other negative sentinels" that don't exist.

dchaudhari7177:
Thanks for the precise diagnosis — both confirmed on my side, and both fixed in 5438104.

**1. Popup reopening by itself.** Reproduced exactly as described. `diffExpanded` now clears in both `select()` and `deselect()`, mirroring `detailCtrl.deselect()`:

```ts
select(repo: string) {
  ...
  this.diffExpanded = false;
  ...
}

deselect() {
  this.selected = false;
  this.diffExpanded = false;
}
```

Added three regression tests for it — `deselect()` closing the popup, `select()` closing one left open by a previous selection, and the full band → commit-row → band gesture from your repro. The comment above them records *why* the reset is needed at all (the component-local `$state(false)` used to get it free on remount), so the next person doesn't delete it as redundant.

**2. CRLF.** Confirmed, and thanks for tracking down *why* — I would not have found the NUL/`text=auto` interaction on my own. Re-committed with LF:

```
git diff --stat origin/dev...HEAD
 src/islands/workdir/Workdir.svelte         | 14 ++++---
 src/islands/workdir/workdir.svelte.test.ts | 63 ++++++++++++++++++++++++++++++
 src/islands/workdir/workdir.svelte.ts      | 16 ++++++++
 src/legacy/main.ts                         |  8 +++-
 4 files changed, 94 insertions(+), 7 deletions(-)
```

`workdir.svelte.ts` is +16 rather than 2498-changed, so blame stays intact.

**3.** Left both notes alone as you said — the empty-pane follow-up and the `.gitattributes` fix are yours, and I did not want to collide with either. Happy to drop the unreachable `hit.row < 0` guard in a follow-up if you'd rather it go.

Full suite green locally: 45 files, 1415 tests.

zangjiucheng (approved):
Both fixed, both verified here.

**CRLF.** All four files are pure LF now, and blame is what actually matters — 16 of 1258 lines attributed to this commit, the rest still spread back across the original working-directory commits:

```
$ git blame HEAD -- src/islands/workdir/workdir.svelte.ts | ...
  97  Add working-directory view: stage/unstage, commit, discard, and stash
  84  Add hunk/line-level staging (git add -p equivalent)
  28  Render staged/unstaged Workdir files as a folder tree instead of a flat list
  24  Add a git blame (line-annotation) view
   …
```

**Popup reset.** Correct in both spots, and the ordering in `main.ts` still works out — `selectWorkdir()` clears the flag synchronously and `expandDiff()` sets it right after, with `refreshStatus()` fire-and-forget in between and never touching it. The three regression tests are the right three; the comment above them explaining *why* the reset isn't redundant is the part that'll save someone later.

I re-ran the gate against `dev` as it is now, which picked up #28 in the meantime — that one rewrote a lot of `main.ts`, so the merge was worth checking rather than assuming. Clean, and both branches' work survives intact:

- merged tree: **1431 tests / 45 files**, svelte-check **534 files 0/0**, **21/21** Playwright
- PR head on its own: 1415 tests, 534 files 0/0

Approving.

One thing I'll flag since I'm the one who asked for it: the `select()` half of the reset is belt-and-braces rather than load-bearing — every exit from the workdir view already routes through `workdirCtrl.deselect()` (`main.ts` 1835, 1876, 2970, 3350), so the flag is always false by the time `select()` runs.

https://github.com/zangjiucheng/GitCat/blob/5438104faa339ee5a0584f12297ca512913cd971/src/islands/workdir/workdir.svelte.ts#L376-L395

Where it isn't free: a full reload with the workdir row selected re-enters through `pendingReselect={workdir:true}` → `selectWorkdir()`, so pulling or pushing while the expanded diff is open now closes it. Defensible — the tree may well have moved under you — but it's a behaviour change I asked for, not one you chose, so it's mine to keep or revisit.

Leave the unreachable `hit.row < 0` guard where it is. It costs nothing and it's the kind of thing that stops being unreachable the moment someone adds a third sentinel.

Thanks for the clean turnaround on this. 🐱

PR #51, opened 2026-08-18
i18n: add a Korean (ko) locale

JerryHyun:
Adds a Korean (`ko`) locale. Additive by design: 60 files under `src/i18n/locales/ko/`, plus the two-line registration `CONTRIBUTING.md` asks for. English stays the source of truth, and anything missing from `ko` falls back to it.

I don't expect you to read Korean — the bottom of this description says what I'd find most useful from a review, and what I verified so you don't have to take my word for it.

## What's here

| | |
|---|---|
| Locale files | 60 namespaces, ~2,050 keys — every namespace `en/` has |
| Registration | `Locale` union, `LOCALES`, and `readStored()`'s guard in `src/i18n/i18n.svelte.ts` |
| `index.html` | **Untouched** — see Fonts below |
| App code | One line in `src/legacy/main.ts` — see below |

Both language pickers (Settings and the setup wizard) render from `LOCALES`, so Korean appears in them with no extra wiring, and `import.meta.glob` picks the folder up on its own.

## The one app-code line

`op.long` already read its label through `t("legacy.op_working")`, but `op.progress` — the event that fires repeatedly right after — had a hardcoded `"working"`. The label showed `处理中 · 0 / 10000`, then flipped to `working · 5 / 10000` and stayed English for the rest of the operation. **This affects Chinese exactly the same way** and is present on `dev`; the key already existed in all three locales, so the fix is one line and adds no strings.

It's in its own commit (`fe3a179`), so it can be lifted out into a separate PR without disturbing the locale files if you'd rather this one stayed locale-only.

## Fonts — measured, and the answer was "do nothing"

`CONTRIBUTING.md`'s step 3 says to add font fallbacks "if the script needs it". I planned to, then measured it in this project's own Playwright browser (Chromium, Windows 11) by rendering into a canvas and comparing widths and ink pixels:

| stack | width of 한글 테스트 | ink hash |
|---|---|---|
| current `--ui` | 94.93 | `1321926971` |
| `--ui` + `"Malgun Gothic"` | 94.93 | `1321926971` |
| `"Malgun Gothic"` alone | 96.33 | `3192064289` |

The first two are byte-identical: Blink's per-character fallback resolves Hangul before the appended name is ever consulted, so the append is a no-op — it doesn't even reach the font, as the third row shows. A `--mono` argument about column alignment turned out backwards too: Hangul is already uniform at exactly the font size under every stack, including the generic one.

So `index.html` ships unchanged. Two caveats I'd rather state than bury: this was measured on Windows only, and it means Korean rendering depends on the OS's own substitution rather than a named family. If you'd prefer the explicit fallbacks anyway for consistency with the Chinese ones, that's three lines in `index.html`.

## ~~Seven keys that belong to an unmerged PR~~ — resolved

> **#28 has merged, so this section is moot: parity is now exact in both directions.** Left in place for the record.

`ko/sidebar.ts` carries seven keys that only exist because #28 introduces them (`checkout_remote`, `checkout_named`, and five `jump_*`). They're in two separate commits (`31262ea`, `fe8ab2b`) so they're easy to find.

If #28 doesn't land, they're harmless — nothing looks them up, so they'd just be unused entries to delete later. If #28 lands first, they're already correct. Either way this PR doesn't depend on that one.

## Terminology

The rule I followed: use the established Korean word where one genuinely exists, and the transliterated English term where that's what Korean developers actually say. So 병합 / 원격 / 브랜치, but 체리픽 / 리베이스 / 스태시 — a Korean developer reading "선별 적용" would have to stop and work out that it means cherry-pick.

Labels are noun-form (`새 브랜치`), messages are the polite conversational register that matches Tama's voice in English (`이제 {name}에 있어요`). Tama's Japanese interjections are kept verbatim, and the 19 strings where English has `にゃ〜` get an equivalent Korean cat-speech ending.

## What I verified

- **Key parity:** every `en` key has a `ko` entry — no silent English fallbacks.
- **Placeholder parity:** every `{name}`, `{n}`, `{sha}` … is present and spelled identically in its Korean counterpart. This one I checked mechanically with an AST script rather than by eye, because a dropped placeholder is invisible until it's on screen.
- **Embedded markup:** the `{@html}` strings keep their tags balanced, and the accent spans wrap the word they're meant to accent.
- **Destructive copy:** force-push, `reset --hard`, discard, stash drop, branch delete and filter-repo read against the English line by line. Nothing softened, nothing inverted, no promise of a recovery path that doesn't exist.
- **Call sites, not just keys:** strings assembled from fragments (`_pre`/`_bold`/`_post`) were reassembled at their render site and read as whole sentences. Korean particles change shape depending on the preceding word, so a per-key read isn't enough.

`svelte-check` clean · vitest 1440/1440 · `vite build` clean — in the default parallel pool.

*(An earlier version of this description warned that the suite needed `--no-file-parallelism`. That is no longer true on the merged tree, so please ignore it.)*

## One known rough edge

`sidebar.force_switch_arm` interpolates a count that the caller builds as an English string (`"3 files"`, at `sidebar.svelte.ts`), so the Korean sentence has an English fragment in the middle of it. **Chinese has this too** — it's not a Korean regression. Fixing it means changing the call site and the `en`/`zh` strings, which felt out of place in a locale PR.

## What would help most from a review

Since the Korean itself isn't checkable for you, the things worth your eyes are:

1. **Is the one `main.ts` line welcome here, or should it be its own PR?**
2. **Do you want the explicit font fallbacks anyway**, for consistency with the Chinese ones, despite the measurement?
3. **The seven #28 keys** — leave them, or should I strip them until that PR lands?

And if a Korean speaker turns up later and wants to change wording, that's entirely fine by me — I'd rather the locale be owned by whoever uses it than be precious about my choices.


zangjiucheng:
**Up front: this review was done with Claude, and I don't read Korean.** I ran the locale through eight AI reviewers over all 60 namespaces, each finding handed to a separate verifier told to refute it, plus a pass looking only for cross-namespace drift. 48 raised, 9 survived. Treat every Korean judgement below as a suggestion from a machine, not a verdict from a speaker — I can't vouch for any of it, and where you disagree, you're the one who's actually read the language. **None of this blocks the PR.**

Gate on the merge with current `dev`: **1440 tests / 47 files**, svelte-check **607 files 0/0**, build clean — and clean without `--no-file-parallelism` too, so that caveat looks local. Key parity 0 missing, placeholder parity 0 mismatches, all 60 files pure LF.

## Your three questions

**1. The `main.ts` line — keep it here.** Correct and symmetric with `op.long`. But `op.long` / `op.progress` / `op.done` have no emitters: every `.event()` call site passes a literal name and none is an `op.*`, and the plugin allowlist maps to `set(state)`, not `event(name)`. So it's a real bug but a dormant one — the sequence you describe can't currently happen. Doesn't change my answer; I'd just rather you knew than wondered later why it won't reproduce.

**2. Fonts — right call, and here's firmer ground for it.** PingFang SC and Hiragino Sans GB contain **0 of 11172 Hangul syllables**, so nothing in `--ui`/`--display` can capture Korean; it falls through to the system Korean face. Ship `index.html` unchanged.

Your Windows measurement can't quite reach that conclusion on its own, though: appending a name at the *end* of the stack is inert whether system fallback handled the Hangul or whether `"Microsoft YaHei"` supplied it, so identical ink hashes are what you'd see either way. `CSS.getPlatformFontsForNode` over CDP names the family that actually rendered, in the harness you already have. And the fix, if there ever is one, is to put a Korean font *before* the Chinese ones rather than after.

**3. The seven `#28` keys — moot.** #28 merged; post-merge parity is exact both directions.

## The one finding that doesn't depend on reading Korean

**`help.branches_p` describes the wrong gesture.** This one is checkable from the code rather than the language: the English says a single sidebar click jumps the graph and that checkout needs a double-click, and the Korean string is simply missing that entire clause — `Sidebar.svelte`'s `onclick` calls `jumpToRef` and bails on `e.detail > 1`, while `ondblclick` opens the checkout confirm. `zh` has both clauses.

https://github.com/zangjiucheng/GitCat/blob/fe3a179ab0fd296da73f2b21facaed5fa4481110/src/i18n/locales/ko/help.ts#L16-L17

It's the first page a new user reads, so it's the one I'd most like fixed — but if you'd rather do it in a follow-up, that's fine by me.

Claude's suggestion: `… 왼쪽 <b>사이드바</b>는 모든 브랜치와 태그를 나열해요 — 하나를 클릭하면 그래프가 그 마지막 커밋으로 이동하고, 브랜치는 더블클릭(또는 우클릭, 또는 ⋮ 버튼)해야 체크아웃돼요. 체크박스로는 그래프에서 브랜치를 표시하거나 숨길 수 있어요.`

## Claude's other suggestions — your call entirely

These rest on the model's reading of Korean, so please overrule any of them freely:

- **`resolver.stash_didnt_clear`** — English says a submodule staged *without* a matching `.gitmodules` entry; Claude reads `".gitmodules 항목과 맞지 않는"` as "an entry exists and disagrees". If that's right it matters, since it's the sentence telling the user where to look. → `대응하는 .gitmodules 항목 없이 서브모듈이 스테이징된 경우예요`
- **`detail.click_to_copy`** — `"클릭하면 복사돼요"` drops "the full hash". The tooltip sits on the *short* sha chip and `copySha()` copies the full 40-char oid, so that detail is the whole point. → `클릭하면 전체 해시가 복사돼요`
- **`err_tools.tool_name_charset`** — the validator is `is_ascii_alphanumeric()`, but `문자` reportedly means "character" rather than "letter", making the rule near-vacuous. → `영문자`
- **`filterrepo`** — "rewrite history" is `재작성` in the title and buttons but `다시 쓰기` almost everywhere else in the same file, and the two show up on screen together mid-run. Unifying on `재작성` would match the menu entry.
- **Five completion toasts** — `multimerge.merge_cheer` / `octopus_complete` / `sequential_complete` and `rebaseplan.cheer` / `complete` are bare `…완료.` while `resolver.rebase.cheer` renders the same English as `리베이스를 완료했어요.`, so Tama's voice changes depending on the dialog.
- **`sidebar.create_failed`** — the one `*_failed` key that drops the noun; every sibling is `<동작>에 실패했어요 — {error}`. Same in `commitmenu.err_create`. → `생성에 실패했어요`
- **Three one-off splits** — `workdir.clean` is `정상` while prose says `깨끗`; `rerere.title` says `해결 상태` while the rest of the file says `해결 방법`; `plumbing.subtitle` renders "tag" twice, once as `tag` and once as `태그`, and leaves `branch` English.

There are a handful of smaller ones too (a dangling `위치는` in `terminal.demo_pre`, `해야 해요` pulled inside a `<b>`, `라벨` vs `레이블`) — I can paste them if useful, but nothing renders wrong.

And your closing point stands on its own: if a Korean speaker turns up later and wants to change any of this, that's the right way for it to get settled — better than a machine's opinion or mine.

## Not yours, but this PR is how I found it

**The type-to-arm danger dialog is entirely untranslated.** All six `armDanger` calls in `sidebar.svelte.ts` — deinit/remove submodule, force switch, delete branch, reset to upstream, delete tag — pass **zero** translated fields. Title, description, "What happens" list, backup note, confirm button: all hardcoded English.

https://github.com/zangjiucheng/GitCat/blob/3be19ecccf7195fc6d89de22d3edb7d256271828/src/islands/sidebar/sidebar.svelte.ts#L1973-L1986

And the shared arm line is English at *all ten* call sites, `resethead` and `forcepush` included, because `armDanger` builds it from literals:

https://github.com/zangjiucheng/GitCat/blob/3be19ecccf7195fc6d89de22d3edb7d256271828/src/legacy/main.ts#L2403-L2407

`filterrepo.danger_type_label` exists in `en`, `zh` and now `ko`, and nothing references it.

So a Korean user deleting a branch gets Tama warning them in Korean over a full-screen English modal. Chinese users have had this the whole time. Mine to fix in app code — but it means "Korean support" isn't finished when this merges, and I'd rather say so than let you hear it from a user.

One last thing: `#50` added a `preview` namespace after your branch point — 22 keys with no `ko/`, falling back to English. Fine by policy. It does mean your test's comment ("no key present in `en/` and genuinely absent from `ko/`") stops being true on merge, and `preview.before` becomes a real key for the `ko -> en` hop it says it can't reach.

Nice work — the register consistency across 2054 keys is what stood out most, at least to a machine that counted: zero `습니다` slips, and `…다옹` landing on exactly the 19 strings that have `にゃ〜`. 🐱

zangjiucheng (approved):

zangjiucheng:
Let me know if you have any comments; I won't block this merge. Thanks for the translations.

JerryHyun:
Thank you — and thank you especially for saying up front how the review was produced. Knowing which parts are a machine's reading and which are checkable from the code is what made it possible to act on, and it's a better way to hand over a review of a language you don't read than pretending to a confidence you don't have.

Pushed as `4a7ba36`, which also merges current `dev`.

## Three things where you were right and I wasn't

**My commit message for the `main.ts` line describes something that can't happen.** I wrote that a Korean user "watched the label flip to English mid-operation" — but as you found, nothing emits `op.*`, so nobody has watched anything. I wrote it as an observation when it was an inference from reading two adjacent lines. The fix is still right; the sentence justifying it wasn't. Thanks for flagging it rather than letting it stand.

**My font measurement doesn't prove what I claimed.** You're right that appending a name at the *end* of the stack is inert either way, so identical ink hashes can't tell "system fallback resolved the Hangul" apart from "Microsoft YaHei supplied it". I reported the measurement accurately but drew a stronger conclusion than it supports. Your evidence — PingFang SC and Hiragino Sans GB containing 0 of 11172 Hangul syllables — actually settles it, and the note that a real fix would have to go *before* the Chinese names rather than after is the part I'd have got wrong if I'd ever gone back to it. Noted in the branch so the next person doesn't repeat my version.

**The `--no-file-parallelism` caveat is stale.** I've just run the merged tree in the default pool: 1440/1440. Please ignore that paragraph in the description — I'll delete it.

## Fixed

**`help.branches_p`** — you were right that this was the one to care about. My branch forked before #28 landed, so `en` and `zh` got the new sentence and Korean kept describing the old gesture, on the first page a new user reads. Fixed, close to your suggested wording.

**`ko/preview.ts`** — added, 22 keys, so parity is exact both directions again.

**The parity comment in the test** — you're right that it was asserting an invariant that had already quietly become false. It now says what's true today and points out that CONTRIBUTOR POLICY lets a locale lag on purpose, which is the whole reason the `ko -> en` hop exists.

I checked each of the Korean suggestions against the English rather than taking them on faith, and took most of them:

- **`stash_didnt_clear`** — correct, and it mattered. English says there's no matching `.gitmodules` entry; my Korean said an entry existed and disagreed. That's the sentence telling someone where to look.
- **`detail.click_to_copy`** — correct. The tooltip sits on the *short* sha chip and the handler copies the full 40-char oid, so "the full hash" was the whole point of the sentence and I'd dropped it.
- **`err_tools.tool_name_charset`** — correct against `is_ascii_alphanumeric()`.
- **`filterrepo`**, **the five completion toasts**, **`create_failed`**, and the `clean` / `rerere` / `plumbing` one-offs — all real, all taken. The five toasts were the most interesting: they go through `bridge.cheer` and `tama.say`, so Tama's voice was changing depending on which dialog she was speaking from.

## One I didn't take, and why

**`externaltools.commit_help_bold`** carries the polite ending inside the `<b>`, where English emphasises only the action. Trimming the Korean to match would cut the word `종료해야` in half, and a bold span that opens or closes mid-word is the worse of the two problems — it's the exact defect I fixed in three other places on this branch. Left as is, with the reason in the commit.

## The untranslated danger dialog

Thank you for looking past the diff and finding that. You're right that it means Korean support isn't finished when this merges, and I'd rather that be said out loud in the PR than discovered by a user.

## Separately

I've opened #52 for the non-blocking note you left on #28 — a later jump now supersedes an armed warning, success included. Your repro dropped straight into a spec.

And on the closing point: yes, please. If a Korean speaker turns up and wants to change any of this, that's a better outcome than my version surviving out of politeness.

zangjiucheng:
Re-checked `4a7ba36`: parity exact both directions (2076/2076), 1440/1440 in the default pool, svelte-check 608 files 0/0, CI green. `help.branches_p` reads better split into two sentences than the version I suggested, and declining the `commit_help_bold` change was the right call — `종료해야` is one 어절, so the bold would have opened mid-word.

What I appreciated most was that you checked each Korean suggestion against the English instead of taking a machine's word for it, and that you corrected your own description where the review turned out to be right. That's a better outcome than me being right about any particular string.

The untranslated danger dialog is mine — I'll pick it up separately, and thanks for filing #52.

Merging now. Thank you for the Korean. 🐱

PR #52, opened 2026-08-18
fix(sidebar): a later jump supersedes an armed warning, success included

JerryHyun:
Picks up the non-blocking note you left when approving #28.

The cancel lived inside `warnJumpFailed`, so the two success paths returned without clearing a warning an earlier click had already armed. Click a ref that isn't loaded, then click a different one inside the hold window and land on it — half a second later Tama scolds you about the first ref while you're looking at the second.

It's the same stale-scolding the hold was added to prevent, just across two refs instead of inside one double-click. Moving the cancel to the top of `jumpToRef` makes every outcome supersede whatever was pending, which is what the hold should have meant all along. `warnJumpFailed` keeps its own clear for any caller that doesn't go through `jumpToRef`.

## Tests

Two, and they're not the same kind — worth separating so the second doesn't look like evidence it isn't:

- **`a later SUCCESSFUL jump supersedes an armed warning`** — the regression. I reverted the one-line change and confirmed it fails without it, then put it back.
- **`a later FAILING jump replaces the armed warning rather than adding to it`** — passes with or without this change, because `warnJumpFailed` already cleared. It's there to pin the behaviour if that inner clear is ever removed as redundant, not to prove the fix.

`svelte-check` 0 errors · vitest 1440/1440 (default parallel pool).

Thanks for writing the repro out in the review — it dropped straight into a spec.

### DIFF
--- a/src/islands/sidebar/sidebar.svelte.test.ts
+++ b/src/islands/sidebar/sidebar.svelte.test.ts
@@ -2821,6 +2821,27 @@ describe("jumpToRef (click a sidebar ref -> select its tip commit)", () => {
     expect(bridge.tama.warn).not.toHaveBeenCalled();
   });
 
+  it("a later SUCCESSFUL jump supersedes an armed warning — no scolding about the ref you left", () => {
+    vi.mocked(bridge.goToOid).mockReturnValue(false);
+    sidebarCtrl.jumpToRef("tag", "v0.9.0", "a".repeat(40)); // not loaded -> warning armed
+    vi.advanceTimersByTime(100); // still inside the hold
+    vi.mocked(bridge.goToOid).mockReturnValue(true);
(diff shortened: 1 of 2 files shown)


zangjiucheng (approved):
Exactly the fix, and the comment says why the cancel belongs at the top rather than leaving someone to rediscover it.

Re-verified against `dev` as it is now, which picked up #51 in the meantime — merge clean, diff still just your two files. 1442 tests / 47 files, svelte-check 608 files 0/0, 21/21 Playwright. I reverted the one line rather than taking your word: test 1 fails without it, test 2 passes. Exactly as you described.

One note: test 2 only fails when *both* clears are gone, and test 1 already catches that, so it never fails alone — I checked, deleting the inner clear leaves all 274 sidebar tests green. Keep it anyway; it's the same redundancy I asked you for on #48. Just don't count test 2 as its cover.

Approving. 🐱

Issue #91, opened 2026-08-20
CI has no `timeout-minutes`, so a hung job runs for six hours


zangjiucheng:
`.github/workflows/ci.yml` sets `timeout-minutes` on **no job**, so a hang runs until GitHub's six-hour default kills it, producing no readable diagnosis in the meantime.

### What this actually cost

Cutting v1.2.0, the bisect suite hung. Three attempts on the same PR:

| attempt | stuck in | duration |
|---|---|---|
| 1 | `Test` | 87 min |
| 2 | apt install | 21 min |
| 3 | `Test` | ~4 h |

Roughly 5.5 hours of runner time, and none of it produced a usable signal — the root cause was found by reproducing locally and sampling the hung process, not from CI.

Worse, the failure was *misleading*: two different stall points made it look like GitHub infrastructure rather than a real defect, which cost more time chasing the wrong thing.

### Fix

Add `timeout-minutes` per job. Reference durations from healthy runs: Rust build+test ~7 min, Windows build ~10 min, frontend ~3 min, e2e ~5 min. Two to three times headroom, so roughly `timeout-minutes: 20` on the Rust jobs.

Worth doing on `release.yml` and `nightly.yml` too — a hang during a release build is strictly worse.

This fixes no bug. It converts "hangs for hours, tells you nothing" into "fails in 20 minutes with a log."

### Good first issue

A few lines of YAML, and the reasoning for each number is already written above.

---
<sub>Found during v1.3.0 research: parallel codebase surveys, then each finding independently verified against the code before filing. Line numbers are from `dev` at the time of writing.</sub>

siyametj:
I have already added timeout-minutes to all workflow jobs across ci.yml, docs.yml, nightly.yml, and release.yml in my existing Pull Request. Waiting for it to be reviewed and merged!

PR #93, opened 2026-08-20
docs: fix plugin manager menu path and installation instructions

### LINKED ISSUE #74: Wrong menu path in three places — it is Tools ▸ Plugins…, not Settings
Three places tell users to open the plugin manager from Settings:

- `docs/plugins.md:329`
- `docs/plugins.md:362`
- `examples/plugins/README.md:36`

The actual location is **Tools ▸ Plugins…** (or ⌘K), per `menu.rs:225`. `docs/guide/plugins.md:15` already has it right.

### Also on that README line

The same parenthetical in `examples/plugins/README.md:36` implies you install by picking the plugin's *folder*. You pick the `plugin.json` file — as `docs/plugins.md:330`, `docs/guide/plugins.md`, all four example READMEs and the `en`/`zh`/`ko` strings ("Plugin manifest (plugin.json)") already say.

Fix the text. Do **not** change the picker to folder-only — that would break a documented, translated flow to match one stray parenthetical.

### Good first issue

Three lines, no code.

---

siyametj:
### Summary of Changes 🛠️

- Fixed menu path from `Settings` to `Tools ▶ Plugins...` in `docs/plugins.md` (lines 329, 362).
- Updated installation instructions in `examples/plugins/README.md` (line 36) to refer to `Tools ▶ Plugins...` and removed the parenthetical `(or its folder)`.

Fixes #74

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
   rust:
     name: Rust (build + test)
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     defaults:
       run:
         working-directory: src-tauri
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,6 +208,7 @@ jobs:
   rust-windows:
     name: Rust (Windows build)
     runs-on: windows-latest
+    timeout-minutes: 20
     defaults:
       run:
         working-directory: src-tauri
@@ -243,6 +244,7 @@ jobs:
   frontend:
     name: Frontend (check + build + test)
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -287,6 +289,7 @@ jobs:
   e2e:
     name: E2E (Playwright)
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - name: Checkout
         uses: actions/checkout@v7
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,6 +29,7 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


siyametj:
I have added timeout-minutes to all workflow files as requested!

zangjiucheng (changes requested):
The part #74 actually asked for is right: all three menu paths are fixed, and you took the correct option of leaving the picker alone rather than changing it. The CI timeouts are good work too, and the values hold up against real run history rather than being guessed, the Windows Rust job peaks around 11 minutes against your 20 cap and the release build at 16m50s against 45.

Two things to fix.

`docs/plugins.md:205` has an editing artifact: the blank line before the `### The two manifest fields` heading was replaced with `Tools ▶ **Plugins...** tab →`. With no blank line in front of it VitePress treats it as a lazy continuation, so the rendered page reads "...is a git() function. Tools ▶ Plugins... tab →". Delete that line and restore the blank one.

`docs/plugins.md:12` replaces the intro trust-model paragraph with a generic feature blurb, which drops both the "a plugin can do anything the command you wrote can do" warning and the only top-of-page link to Security & trust. That sentence's "no sandbox" clause is genuinely stale now that Luau handlers exist, so a refresh is welcome, but as a rewrite in its own PR rather than a deletion here.

Two smaller ones: pick a single rendering of the menu path and use it in all three places, `Tools ▸ Plugins…` is what `docs/guide/plugins.md:15` uses. And the button actually reads `＋ Install from file…`, not "Install plugin…", while `examples/plugins/README.md:36` still says "tab" when the plugin manager is a modal.

Last thing, and it is the one that matters most for merging: the workflow timeouts fix #91, not #74, but the body only says "Fixes #74", so merging this closes the wrong issue and leaves #91 open. I would rather you split them, since the docs half can go in as soon as the two fixes above are done, and the YAML half deserves its own review. If you would rather keep them together, retitle and add "Fixes #91" to the body.

PR #94, opened 2026-08-20
Refactor locale loading mechanism

siyametj:
Refactor locale loading to use eager loading for English and lazy loading for other locales. Update the current locale initialization and setLocale function to handle dynamic loading.

### DIFF
--- a/src/i18n/i18n.svelte.ts
+++ b/src/i18n/i18n.svelte.ts
@@ -26,16 +26,13 @@ export const LOCALES: { id: Locale; label: string }[] = [
   { id: "ko", label: "한국어" },
 ];
 
-// ONE glob over every locale dir — `./locales/<loc>/<namespace>.ts` — so adding
-// a language is PURELY ADDITIVE: create `locales/<loc>/`, add the id to `Locale`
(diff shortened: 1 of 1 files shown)


zangjiucheng:
Before either of us puts more work into this: can you measure what it actually buys? A cold-start before and after, on `tauri dev` and on a release build, would settle it. GitCat loads its bundle from local disk over the asset protocol, so this saves parse and eval of roughly 316KB of source rather than a download, and the whole locale set is only about 456KB to begin with.

I should be upfront that I am leaning toward not merging. The cost is turning a synchronous, always-correct API into an asynchronous one, and as written a user with zh or ko saved boots the entire app in English with no recovery: `DICTS` is a plain object rather than `$state`, so the late assignment re-renders nothing, and `setLocale` early-returns because `current` is already the locale whose dictionary never arrived. Doing that properly is real work and I would rather know the payoff first.

If the number is meaningful, say 100ms or more, I am happy to take the idea, though I would want boot to await the saved locale's chunk rather than render English and swap. If it is a few milliseconds, let's close it. Either way, note `src/i18n/i18n.svelte.test.ts:19` asserts `locale()` on the line right after `setLocale()`, so it fails on this branch today.

PR #95, opened 2026-08-20
Update danger type label and confirmation button

siyametj:
Refactor danger type label to use translation functions and improve UI consistency.

### DIFF
--- a/src/legacy/main.ts
+++ b/src/legacy/main.ts
@@ -2402,16 +2402,11 @@ function armDanger(ctx){
   // it's always present.
   const typeLabel=$("#dangerTypeLabel");
   typeLabel.textContent="Type the "+(ctx.typeNoun||"branch name")+" ";
-  const tn=document.createElement("b"); tn.className="mono"; tn.id="dangerTypeName"; tn.textContent=ctx.name;
-  typeLabel.appendChild(tn);
-  typeLabel.appendChild(document.createTextNode(" to "+(ctx.typeVerb||"arm the rewrite")+":"));
-  const inp=$("#confirmInput"); inp.placeholder=ctx.name; inp.value="";
-  $("#dangerGo").textContent=ctx.confirmLabel||"Confirm"; $("#dangerGo").disabled=true;
-  openScrim("#dangerScrim"); setTimeout(()=>inp.focus(),30);
-}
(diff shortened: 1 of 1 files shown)


zangjiucheng (changes requested):
This one needs to go back, and I want to be clear it is a mechanical accident rather than a wrong idea.

The patch deletes the `}` that closes `armDanger()`, so everything from `let dangerBusy=false;` to the `export { … }` at the end of `main.ts` ends up inside the function body and the module is a hard SyntaxError. The same hunk removes `disarmDanger()` while three call sites remain, including the global Escape handler, and it removes the `#confirmInput` listener that is the only code that ever enables `#dangerGo`, the `#dangerCancel` listener, and the `openScrim("#dangerScrim")` call. So the destructive-action modal never opens, and if it did, the type-the-name gate protecting every destructive action would be gone.

The underlying problem is real: that label is hardcoded English on dev and should be translatable. Two things to get right in the redo.

`filterrepo.danger_type_label` is the static `data-i18n-html` placeholder that `index.html` consumes, not a runtime string. It has no `{noun}`/`{verb}` slots, it hardcodes the name `main`, and it contains markup that `textContent` renders as visible raw tags. Add new runtime keys instead and keep building the `<b class="mono">` node between them, exactly as dev does:

```ts
// src/i18n/locales/en/filterrepo.ts
danger_type_prefix: "Type the {noun} ",
danger_type_suffix: " to {verb}:",
```

And keep the contract one-directional: `ctx.typeNoun`, `ctx.typeVerb` and `ctx.confirmLabel` are display strings that callers have already translated, so do not wrap them in `t()` inside `armDanger`. Translating a display string only appears to work because `t()` falls back to its argument. `armDanger` should translate its own frame and nothing else.

Every new key also has to exist in `src/i18n/locales/en/` in the same commit. `sidebar.branch_name`, `sidebar.arm_rewrite` and `common.confirm` are in no locale at all, so the confirm button would literally read `common.confirm`.

Please run `pnpm check && pnpm build && pnpm test` locally before pushing, as CONTRIBUTING.md asks. Fork PRs need a maintainer to approve the workflow run, so nothing but GitGuardian has run on your PRs and you would have had no way to see any of this.

PR #96, opened 2026-08-20
i18n: key-ify danger dialog texts and use placeholders for locales

### LINKED ISSUE #92: 63 MB of superseded demo video is still in git history
v1.2.0 moved `docs/public/demo.{mp4,webm}` into Git LFS, which stops future re-records from adding pack weight. It does **not** shrink existing clones — the old blobs are still in history.

### The measurement

The pack is roughly 127 MB. Video accounts for **7 blobs / 63 MB**, and video does not delta-compress, so that is close to half the clone cost for a single asset:

```
15.2 MB  docs/public/demo.mp4       <- current
15.2 MB  docs/public/demo.webm      <- current
13.7 MB  docs/public/demo-4k.mp4    <- deleted long ago
 6.1 MB  docs/public/demo.mp4       <- superseded
 5.0 MB  docs/public/demo.mp4       <- superseded
 4.2 MB  docs/public/demo.webm      <- superseded
 3.7 MB  docs/public/demo.webm      <- superseded
```

**More than half of it (~33 MB) is files that exist on no branch.**

### What it takes

```bash
git lfs migrate import --everything --include='*.mp4,*.webm'
```

then force-push every branch.

### Why it is not simply done

That rewrites every commit SHA on every branch. It breaks open PRs and every fork, and everyone with a clone must re-clone or hard-reset. `main` and `dev` also carry a `non_fast_forward` rule, so branch rulesets need temporary changes.

This wants an announcement window and a quiet moment with no open PRs — it is a coordinated decision, deliberately not a solo one.

### Alternative worth weighing first

The repo already has this pattern: `scripts/fetch-pdfium.mjs` fetches a binary at postinstall rather than committing it. The demo video could live on a GitHub Release and be fetched during the docs build — no LFS quota exposure, no history rewrite for future assets. It would not reclaim the 63 MB already in history.

---

siyametj:
### Summary of Changes
- Replaced hardcoded English text in `src/legacy/main.ts` with `t()` translation keys and placeholders (`noun` and `verb`).
- Refactored `deleteBranch` and `removeSubmodule` in `src/islands/sidebar/sidebar.svelte.ts` to use localized string keys instead of direct string concatenation.
- Ensures destructive action modal prompts render properly across non-English locales (e.g., Korean and Chinese).

Fixes #92

### DIFF
--- a/src/islands/sidebar/sidebar.svelte.ts
+++ b/src/islands/sidebar/sidebar.svelte.ts
@@ -40,16 +40,16 @@
 // to refresh" precedent below.
 //
 // Submodules (M4 — deinit + remove, on top of M1-M3's status/init/add/sync):
-// deinitSubmodule/removeSubmodule are the per-row destructive actions,
+// /removeSubmodule are the per-row destructive actions,
 // routing through the shared armDanger typed-confirm scrim exactly like
-// deleteBranch/deleteTag above. deinitSubmodule is status-gated the same way
+// deleteBranch/deleteTag above.  is status-gated the same way
 // doDeleteBranch's own "isCurrent" checks are: submoduleNeedsForceConfirm(
 // status) mirrors real git's own precondition (a dirty tree OR a merge-
 // conflicted gitlink both refuse `deinit` without `-f` — see submodule.rs's
 // module doc comment) — everything else (clean/out-of-date/not-initialized)
 // calls straight through with force:false, no scrim at all, matching this
 // app's "never show a needless confirm for a safe operation" rule. Its
-// doDeinitSubmodule private helper has the same two-tier fallback as
+// do private helper has the same two-tier fallback as
 // doDeleteBranch: a plain force:false attempt first, then (only for the
 // stale-status race where a row looked safe but git itself refuses) a
 // window.confirm()-gated retry with force:true. removeSubmodule always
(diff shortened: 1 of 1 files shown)


zangjiucheng (changes requested):
The goal is right and the mechanism you picked is the correct one, `forcepush.svelte.ts:87` already does interpolation this way. But a case-insensitive find-and-replace of `deinitsubmodule` went through code this PR had no reason to touch, and the result does not parse.

`sidebar.svelte.ts:1432` is now `async (path: string, status: string) {`, which is a hard SyntaxError, so the bundle, `svelte-check` and vitest all fail. `doDeinitSubmodule` became `do`, the public `deinitSubmodule` entry point is gone while `Sidebar.svelte:1125` still calls it from the Deinit button, and the header comments at lines 43, 45, 52, 61 and 1551 got mangled by the same pass.

```ts
async deinitSubmodule(path: string, status: string) {          // 1432
private async doDeinitSubmodule(path: string, force: boolean) { // 1461
```

Second, none of the roughly 17 new keys exist anywhere, including `src/i18n/locales/en/`, so every converted string renders as its raw dotted key and the interpolated path vanishes with it. English is the source of truth (CONTRIBUTING.md); zh and ko are allowed to lag, en is not. And `confirmLabel`, `typeNoun` and `typeVerb` are handed to `armDanger` as bare key literals that never go through `t()`, so the prompt would read "Type the sidebar.submodule_noun vendor/a to sidebar.remove_verb:". `armDanger` takes display strings, so translate at the call site:

```ts
// src/i18n/locales/en/sidebar.ts
remove_submodule_title: "Remove submodule — {path}",
remove_submodule_verb: "Remove submodule",
// call site
confirmLabel: t("sidebar.remove_submodule_verb"),
```

One shape suggestion: rather than splitting the `lose` block into one key per `<li>` plus a shared `common.what_happens`, copy the forcepush shape and keep one `lose` key holding the whole `<h5>…</h5><ul>…</ul>` with placeholders. Translators need whole sentences; fragments reassembled in English word order do not survive most languages.

`sidebar.svelte.test.ts` also asserts on the strings you changed, so it needs updating in the same commit. Please run `pnpm check && pnpm build && pnpm test` before pushing. Fork PRs need a maintainer to approve the workflow run, so nothing except GitGuardian has run on any of your PRs so far, which is why none of this showed up as a red check.

PR #97, opened 2026-08-20
docs: document declarative panels in plugin manifest

siyametj:
### Summary of Changes
- Documented the `panels` section in `docs/plugins.md` for plugin authors.
- Updated top-level manifest table and intro bullet points to include declarative side panels.
- Added detailed documentation for panel fields (`id`, `title`, `items`) and all supported widget item types (`Text`, `Heading`, `Button`, `CommandOutput`).
- Included a complete, annotated JSON example showing declarative panel integration.

### Context
Declarative panels are fully implemented in `plugin_registry.rs`, but were previously omitted from the plugin documentation.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,12 +1,13 @@
 ---
-description: Author a GitCat plugin — a declarative plugin.json that contributes ⌘K commands and lifecycle hooks backed by your own external commands.
+description: Author a GitCat plugin — a declarative plugin.json that contributes ⌘K commands, UI panels, and lifecycle hooks backed by your own external commands.
 ---
 
 # Writing a plugin
 
-A GitCat plugin is a small, **declarative** manifest — a single `plugin.json` file. It doesn't ship code that runs inside GitCat; it contributes two things:
+A GitCat plugin is a small, **declarative** manifest — a single `plugin.json` file. It doesn't ship code that runs inside GitCat; it contributes three things:
 
 - **Commands** — actions that show up in the ⌘K command palette and run an external command you specify.
+- **Panels** — declarative sidebar/UI widgets (headings, text, buttons, command outputs) rendered directly by GitCat.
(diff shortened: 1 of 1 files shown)


zangjiucheng (changes requested):
Thanks for this, and sorry for the slow reply. Panels genuinely are undocumented while `docs/guide/plugins.md` promises they are covered, so this is the right page to fix. The field-level facts check out against `plugin_registry.rs`, and the validation-rule paragraph is the single most useful thing a panel author needs.

Three things before it can land.

The widget type is wrong. `PanelItem` is `#[serde(tag = "type", rename_all = "kebab-case")]` (`plugin_registry.rs:251`), so the only accepted discriminator is `command-output`. An unknown internally-tagged variant fails the whole manifest parse, so a reader copying the example at line 52 gets the install rejected outright.

```json
{ "type": "command-output", "command": "status", "label": "Status" }
```

That is lines 52, 147 and 151.

The section never says how a panel is actually opened. It calls them "sidebar/UI widgets", but `pluginpanels.svelte.ts` `build()` turns each declared panel into exactly one command-palette entry labelled by its `title`, and `PluginPanel.svelte` renders it as a modal. A reader will go looking in the sidebar for something that only exists in the palette. Worth saying that the `title` is what users search for, and that panel-invoked commands only ever get `{repo}` filled, never `{sha}` or `{file}`.

Last one, and the reason I cannot just merge this: the whole-file restyle deletes 16 of the page's 17 markdown links, including the outbound Luau link and the GitHub link to `examples/plugins/`, and strips inline-code backticks across sections that have nothing to do with panels. `docs/` is a deployed site, so that is a live regression. Please rebase the panel content onto the current `docs/plugins.md` and drop the restyle. If you do want the page restyled, open that separately and I will look at it on its own merits.

One process note: this targets `main`, while everything else goes to `dev` first. Please retarget it when you rebase, the Edit button next to the PR title lets you change the base branch.

PR #98, opened 2026-08-21
Make the Rust test suite runnable on Windows

JerryHyun:
`cargo test` did not fail on Windows — it never started. Two commits, and the second one is only reachable because of the first.

```
before:   0 passed,  0 failed
after:  682 passed, 19 failed
```

## 1. Nothing could run

Two separate things, and the aggregate hid both:

- **`tests/submodule.rs` did not compile.** Two tests used `std::os::unix` APIs unguarded. One non-compiling target makes `cargo test --no-fail-fast` emit no test results at all, which is why the whole command reported nothing rather than a partial run.
- **The library unit-test binary did not start.** It exited with `STATUS_ENTRYPOINT_NOT_FOUND` (0xC0000139) before the harness ran anything — taking `export_bindings`, the test that regenerates `src/ipc/bindings.ts`, down with it.

The second one is the interesting half. That binary imports `comctl32!TaskDialogIndirect`, which exists only in comctl32 **version 6**. `C:\Windows\System32\comctl32.dll` is v5 and does not export it; v6 lives in a side-by-side assembly the loader binds only for a process whose manifest declares `Microsoft.Windows.Common-Controls` 6.0.0.0. `tauri_build` embeds that manifest in the app binary, and nothing embeds it in a test binary.

Worth stating precisely, because it is easy to overstate: **exactly two binaries in the build carry comctl32 imports** — the app and the lib unit-test exe. `dumpbin /DEPENDENTS` over `target/debug/deps/*.exe` shows none of the 36 `tests/*.rs` binaries reference comctl32 at all, and a pre-change integration exe still on disk runs fine. The import comes from `tauri-runtime-wry`'s own `dialog/windows.rs`, compiled unconditionally on Windows, and from `muda`/`rfd` under the dialog plugin — so no feature flag drops it.

**Fixed by delay-loading comctl32** for `windows-msvc` targets. There is no way to reach only the test binaries, and all three alternatives were measured rather than assumed — the reasons are in `build.rs` so nobody has to repeat them:

| attempt | result |
|---|---|
| `cargo:rustc-link-arg-tests` | does not apply to the lib's own unit-test target. With it set, the built test exe carries no manifest at all; it covers `tests/*.rs` only |
| unscoped `cargo:rustc-link-arg` | reaches the test binary, but also the app, where `/MANIFEST:EMBED` collides with tauri-build's own manifest resource: `CVT1100: duplicate resource` then `LNK1123` |
| side-by-side `<exe>.manifest` | a build script cannot place it — it does not know the test binary's hashed filename |

The duplicate, not manifest embedding as such, is what blocks that route. `WindowsAttributes::new_without_app_manifest()` would also work by handing `build.rs` ownership of the app manifest; that is noted as the road not taken, with why.

**The app still binds v6** — verified by launching the built app and reading `Process.Modules`, where comctl32 resolves from the WinSxS 6.0 assembly. The cost is fail-fast becoming fail-late: if the app's manifest were ever lost, the old failure was a loader error at startup, and now it is an unhandled structured exception when a dialog opens. That trade is written down next to the flag rather than left for someone to discover.

## 2. The host's `core.autocrlf` decided whether tests passed

With the suite finally running, 88 failed. **69 of them were one setting.**

Git for Windows installs `core.autocrlf=true` by default, which rewrites LF to CRLF on every checkout. A test writes `"base\nline2\n"`, the code under test does something that re-checkouts the file, and the assertion compares against `"base\r\nline2\r\n"`. That was 58. Twelve more were the same cause wearing a different hat — `is_clean()` returning false, because git reports a file as modified when its line endings no longer match the index.

The fixture already takes some care to keep the host's config out of these repos (gpgsign, identity, gc), and `git()` neutralizes global/system config for its own invocations. What it could not cover is **the code under test**, which shells out to `git` itself with no such environment. So the setting has to live in the repo, where local config outranks the host's global.

`core.eol` as well as `core.autocrlf`: autocrlf alone leaves `eol` to decide the checkout form for paths a `.gitattributes` file marks as `text`, and some tests write one. Both are no-ops on Linux, where this is already the effective default.

`init_bare` deliberately does **not** get this — a bare repo has no working tree, so there is no checkout to convert, and setting it there would leave the next reader working out why.

The list also moved into `TempRepo::apply_test_config`, because `patch.rs`'s `clone_of` had hand-copied four of the five entries and drifted. A clone starts with none of its source's local config, and that stale copy is exactly why three patch tests still failed after the first pass. One list, two callers, no third copy to go stale.

## What is still red, and why it is not in here

Nineteen remain (twenty on some runs — one bisect test is timing-sensitive and flips). They are pre-existing and newly *visible*, not regressions: this is the first time any of them has been allowed to run. Each group has its own cause and its own fix shape, so they are filed separately rather than bundled into this PR:

- **8** (#99) — nested submodule repos inherit the host's `core.autocrlf`. Same root cause as above, but `git submodule add` clones the nested repo *inside the call under test*, so there is no moment to configure it from the test. Needs a different mechanism.
- **4** (#100) — `bisect`'s scripted good/bad command is a shell script; Windows has no shell to run it, so bisect converges on the wrong commit.
- **3** (#101) — three test-side assumptions that only hold on POSIX: a filename containing a tab (Windows cannot create it at all), an assertion comparing a raw path against the backslash-escaped form `git config` actually writes, and a rename target with no uniqueness in its name that nothing ever cleans up.
- **1** (#102) — the built-in terminal's ConPTY round-trip reads nothing from the PTY. Unexplained.

Four more failures on my machine were **not** a repo problem: `git-filter-repo` is not installed locally, and CI installs it with `pipx`. Mentioning it only so the count reconciles.

## CI

The Windows job stays compile-only, but its comment no longer claims the suite cannot compile or start — both of which it said, and neither of which is true any more. Turning the run on belongs with fixing the failures above; flipping it now would just make every PR red.

## Verified

`cargo build` links the app cleanly · `cargo test --lib` 255 passed / 1 failed (the ConPTY test) · `export_bindings` runs and leaves `src/ipc/bindings.ts` unchanged · `cargo clippy --lib` clean on the changed files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,20 +188,24 @@ jobs:
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
   # catches a Windows-only compile break on the PR instead.
   #
-  # Compile-only, on purpose:
+  # Compile-only, on purpose — but the reason changed, so read this before
+  # assuming it still can't run:
   #  - `cargo build` covers the app/library binary.
   #  - `cargo test --lib --no-run` additionally COMPILES the library unit-test
-  #    code on Windows (Windows-clean: the handful of Unix-only cases are
-  #    #[cfg(unix)]-guarded), catching a Windows test-compile break too. `--lib`
-  #    skips the tests/ integration suite (submodule / repo_registry /
-  #    repo_files), which uses unguarded std::os::unix APIs and doesn't compile
-  #    on Windows yet (a separate follow-up).
-  #  - We deliberately do NOT run the tests. The Tauri library's test binary
-  #    transitively links the WebView2/wry native stack and fails to resolve a
-  #    DLL entry point at startup on a stock runner (exit 0xc0000139,
-  #    STATUS_ENTRYPOINT_NOT_FOUND) before any test executes. The cli_shim
-  #    launcher tests are platform-agnostic pure functions already run by the
-  #    Linux `rust` job, so running them here would add nothing regardless.
+  #    code on Windows, catching a Windows test-compile break too.
+  #  - The whole test suite now COMPILES on Windows, and the library test
+  #    binary now STARTS. It used to die at startup with exit 0xc0000139
+  #    (STATUS_ENTRYPOINT_NOT_FOUND) before any test ran, because it imports a
+  #    comctl32 v6 export without the manifest that binds v6; build.rs
+  #    delay-loads comctl32 to fix that, and the two remaining Unix-only cases
+  #    in tests/submodule.rs are #[cfg(unix)]-guarded.
+  #  - We still do NOT run the tests here, and that is now a real choice rather
+  #    than an impossibility. Locally the suite is 613 passed / 88 failed on
+  #    Windows: one library test (a ConPTY round-trip that reads nothing) plus
+  #    an integration cluster concentrated in workdir/merge/rebase/submodule.
+  #    Those are pre-existing Windows gaps that have never been allowed to run
+  #    before, not regressions. Turning the run on is a follow-up that belongs
+  #    with fixing them — flipping it now would just make every PR red.
   # No apt/system-deps step: Windows runners ship the MSVC toolchain and WebView2
   # SDK, and libgit2-sys builds its vendored copy with `cc`, same as on Linux.
   rust-windows:
--- a/src-tauri/build.rs
+++ b/src-tauri/build.rs
@@ -8,5 +8,88 @@ fn main() {
     // reports "Compiling gitcat". Watch the whole icons/ directory so any
     // icon regeneration always re-embeds correctly.
     println!("cargo:rerun-if-changed=icons");
+    delay_load_comctl32_on_windows();
     tauri_build::build()
+}
+
+/// Let the library's unit-test binary START on Windows.
+///
+/// It could not. `cargo test --lib` exited with STATUS_ENTRYPOINT_NOT_FOUND
+/// (0xC0000139) before the harness ran anything, which took `export_bindings`
+/// — the test that regenerates `src/ipc/bindings.ts` — down with it. CI runs
+/// the Rust suite on Linux, so nothing caught it.
+///
+/// Cause: this binary imports `comctl32!TaskDialogIndirect`, which exists only
+/// in comctl32 **version 6**. `C:\Windows\System32\comctl32.dll` is v5 and does
+/// not export it; v6 lives in a side-by-side assembly the loader binds only for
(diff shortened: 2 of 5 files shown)


zangjiucheng (approved):
Thanks for this, and sorry it took me so long to get to it. This one sat far longer than it deserved.

This is a strong PR and it does real root-cause work rather than a workaround. The comctl32 v6 and manifest analysis in `build.rs`, with the three rejected alternatives each recorded next to the exact linker error they produce, is the kind of note that saves the next person a full day. Please keep that comment exactly as it is. Separating the two independent causes was the right instinct too: the aggregate "0 passed, 0 failed" was hiding both a non-compiling target and a test exe dying before the harness ever ran.

Approving. Four follow-ups, none of them blocking:

1. `ci.yml:203-205` still carries commit 1's numbers (613 passed / 88 failed, concentrated in workdir/merge/rebase/submodule). Commit 2 in this same PR took it to 682/19 and the survivors are elsewhere, so as written it sends the next maintainer looking in files where nothing fails any more.
2. `ci.yml:196` claims the whole test suite compiles on Windows, but the job runs only `cargo test --locked --lib --no-run`, which never touches `tests/`. Switching it to `cargo test --locked --no-run` makes that claim enforceable. Otherwise the next unguarded `std::os::unix` call in any tests file puts Windows back to zero with CI green, which is exactly how `submodule.rs` got here. Related: nothing in CI ever starts the lib test binary, so the 0xC0000139 fix has no regression guard at all. `cargo test --locked --lib -- --list` costs seconds and proves the exe loads.
3. `patch.rs` `clone_of` currently works by luck of call-site ordering: the `git clone` at `patch.rs:34` still runs under the host's config and checks out CRLF, and only the `reset --hard` at each call site re-checkouts with the new local setting. Worth moving it to clone time and adding the `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` pair that `submodule.rs`'s `FreshClone::of` already uses.
4. Two nits: `common/mod.rs:93` justifies `core.eol=lf` with "some tests write" a `.gitattributes`, and none in this repo does, so it reads as defensive rather than as a fact. And `patch.rs:240`, `:264` and `:338` set `commit.gpgsign=false` on repos that already got it from `TempRepo::init`.

One thing I want to make explicit rather than discover later: `cargo:rustc-link-arg` is unscoped, so the shipped app on both Windows targets now delay-loads comctl32 to fix a test-only startup problem. You documented that trade and the `new_without_app_manifest` alternative honestly, and CI links everything cleanly, so this is a conscious yes from me rather than something that slipped through.

This merges first, ahead of the rest of your stack.

Issue #99, opened 2026-08-21
Nested submodule repos inherit the host's core.autocrlf, failing 8 tests on Windows


JerryHyun:
Eight `tests/submodule.rs` tests fail on Windows because the *nested* repository a submodule lives in inherits the host's `core.autocrlf`.

This only becomes visible once the Rust suite can run on Windows at all — see #98, which fixes that and closes the same problem for every other suite. The fix there sets `core.autocrlf=false` locally in each throwaway repo, which is enough everywhere except here.

### Why the same fix does not reach these

`git submodule add` clones the child repository into the parent, creating an **independent** git repo. Config is not inherited, so the nested repo starts with none of `TempRepo::apply_test_config`'s settings and falls back to the host's global — which on Windows is `core.autocrlf=true` by default.

And in most of these tests the clone happens *inside the call under test*, so there is no moment for the test to configure the nested repo before its working tree is written:

```rust
// tests/submodule.rs:909
let added = tauri::async_runtime::block_on(submodule_add(...));
assert_eq!(parent.read("sub/f.txt"), "hello\n");   // reads "hello\r\n"
```

### Where

| test | line |
|---|---|
| `submodule_add_clones_new_submodule_and_it_is_immediately_clean` | 909 |
| `submodule_add_with_branch_checks_out_that_branch_not_the_default` | 959 |
| `submodule_deinit_with_force_backs_up_dirty_content_then_clears` | 1225 |
| `bug2_force_deinit_refuses_when_a_nested_submodule_of_a_submodule_is_dirty` | 1684 |
| `submodule_status_on_cyclic_nested_submodule_terminates_cleanly_instead_of_crashing` | 2006 |
| `submodule_update_recursive_handles_nested_submodule_of_a_submodule` | 743 |
| `submodule_update_with_init_clones_and_checks_out_never_initialized_submodule` | — |
| `submodule_update_without_recursive_leaves_the_nested_submodule_uninitialized` | — |

All fail the same way — a compared string differs only by `\r`:

```
assertion `left == right` failed
  left:  "hello\r\n"
  right: "hello\n"
```

### Fix

Per-repo config cannot reach a repo the code under test creates, so this wants a process-wide answer instead: point the test binary's own `GIT_CONFIG_GLOBAL` at a fixture-owned file containing `[core] autocrlf = false`, so **every** `git` the tests or the code under test spawn inherits it — nested clones included.

[...]

PR #103, opened 2026-08-23
Fix three POSIX assumptions in the test suite

### LINKED ISSUE #101: Three tests assume POSIX: a tab in a filename, an unescaped path, a non-unique rename target
Three test-side assumptions that only hold on POSIX. Each is small and independent, grouped because they are the same class of thing: the test, not the code under test, is what assumes a platform.

All three surface only now that the Rust suite can start on Windows — see #98.

### 1. A filename containing a tab — Windows cannot create it at all

`tests/workdir.rs`, `stage_file_accepts_a_tab_containing_filename` (:733)

```
called `Result::unwrap()` on an `Err` value: Os { code: 123, kind: InvalidFilename, ... }
```

A tab is not a legal character in a Windows filename, so the file the test needs cannot exist. The premise does not hold; the assertion is fine. **Fix:** `#[cfg(unix)]` with a line saying why, in the shape `tests/submodule.rs` already uses for its two POSIX-only cases.

### 2. An assertion compares a raw path against the escaped form `git config` writes

`tests/submodule.rs`, `submodule_sync_rewrites_git_config_url_after_gitmodules_is_hand_edited` (:1052)

```rust
parent.must(&["config", "-f", ".gitmodules", "submodule.sub.url", &child_new.path()]);
...
assert!(gitmodules.contains(&child_new.path()), "expected .gitmodules to record the new url");
```

`git config` escapes backslashes when it writes a value, so a Windows path goes in doubled. Reproduced directly:

```
$ git config -f .gitmodules submodule.sub.url 'C:\Temp\GitCat\child'
$ cat .gitmodules
[submodule "sub"]
        url = C:\\Temp\\GitCat\\child
```

`contains(&child_new.path())` looks for the single-backslash form and can never match. **Fix:** read the value back with `git config --get -f .gitmodules` (which unescapes) instead of substring-matching the file text — that also makes the assertion about the recorded value rather than about the file's formatting.

### 3. A rename target with no uniqueness, that nothing cleans up

`tests/submodule.rs`, `submodule_deinit_recovers_offline_via_init_and_update` (:1287)

```rust
let moved_away = child.dir.with_file_name("submodule_deinit_offline_child_GONE");
std::fs::rename(&child.dir, &moved_away).expect("simulate the origin going away");
```

Every other temp path in the suite carries pid + nanos + a counter. This one is a fixed name, and once the rename succeeds nothing owns the result — `TempRepo::drop` removes `child.dir`, which no longer exists, so the renamed copy leaks. I found a full leaked repo sitting there (42 entries, three read-only git objects).

That leak then poisons every later run, because `rename` onto an existing non-empty directory fails — `ENOTEMPTY` on unix, `ERROR_DIR_NOT_EMPTY` on Windows. CI never sees it: a fresh runner always starts with an empty temp dir.

**Fix:** give the target the same uniqueness as every other temp path, and let a `TempRepo` own it so `Drop` cleans it up.

**Not fully explained:** after deleting the leaked directory (including clearing the read-only attribute git leaves on objects), this test still fails with `ERROR_DIR_NOT_EMPTY` on Windows, while renaming an unrelated git repo directory on the same machine works. So there is something else about the repo's state at that moment. The fixed name is a real bug regardless, but fixing it may not be sufficient here.

---
<sub>Found while making `cargo test` runnable on Windows (#98). Each of the three was reproduced directly rather than inferred from the failure text.</sub>

JerryHyun:
Closes #101.

**Stacked on #98** — these tests cannot run on Windows without it, so the diff will show that PR's two commits until it merges. Only `0b55028` belongs to this one.

19 → 18 failing on Windows, and one test stops poisoning every later run on the same machine.

## A tab in a filename

`stage_file_accepts_a_tab_containing_filename` cannot build its input on Windows: a tab is not a legal filename character there, so `std::fs::write` fails with `ERROR_INVALID_NAME` before the assertion is reached. The behaviour under test — that `stage_file` accepts a path git itself accepts — is unchanged on Windows; there is just no way to construct the case.

`#[cfg(unix)]`, in the same shape as the two POSIX-only cases already gated in `tests/submodule.rs`.

## An assertion comparing a raw path to what `git config` actually writes

`submodule_sync_rewrites_git_config_url_after_gitmodules_is_hand_edited` substring-matched `.gitmodules`'s text for a path. `git config` escapes backslashes when it writes a value, so a Windows path lands doubled — reproduced directly:

```
$ git config -f .gitmodules submodule.sub.url 'C:\Temp\GitCat\child'
$ cat .gitmodules
[submodule "sub"]
        url = C:\Temp\GitCat\child
```

`contains(&child_new.path())` looks for the single-backslash form and can never match. It now reads the value back with `git config --get -f`, which unescapes — and which also makes the assertion about the recorded *value* rather than about the file's formatting, which is what the line was trying to say in the first place.

## A rename target with no uniqueness, that nothing cleaned up

`submodule_deinit_recovers_offline_via_init_and_update` renamed the child repo to a **fixed** name to simulate the origin vanishing:

```rust
let moved_away = child.dir.with_file_name("submodule_deinit_offline_child_GONE");
```

Every other temp path in the suite carries pid + nanos + a counter. This one did not — and once the rename succeeded, nothing owned the result: `child`'s `Drop` removes `child.dir`, which by then no longer exists, so the renamed repo leaked into the temp dir.

From that point on, every later run failed, because `rename` onto an existing non-empty directory is an error (`ENOTEMPTY` on unix, `ERROR_DIR_NOT_EMPTY` on Windows). CI never sees it — a fresh runner always starts with an empty temp dir. I found the leaked repo sitting in mine, 42 entries deep.

The target is now derived from the child's own unique directory name, and handed to a `TempRepo` so `Drop` cleans it up.

### Correcting the issue on this one

#101 says the `ERROR_DIR_NOT_EMPTY` was not fully explained, because deleting the leaked directory by hand did not appear to help. With the unique name in place the rename succeeds, so the leftover **was** the whole cause — my manual delete evidently had not taken effect before that run. I would rather correct that here than leave a wrong note standing in the issue.

That test still fails, but now on `"hello\r\n"` vs `"hello\n"`: it has become one of #99's nested-submodule CRLF cases. Different issue, not touched here.

## Verified

On Windows: `--test workdir` 40 passed / 0 failed (was one failing), `--test submodule` 40 passed / 9 failed (was 10). Full suite 682 passed / 18 failed.

Nothing here changes behaviour on unix: the gate only excludes, the `git config` read is equivalent where paths contain no backslashes, and the rename target is merely unique now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,20 +188,24 @@ jobs:
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
   # catches a Windows-only compile break on the PR instead.
   #
-  # Compile-only, on purpose:
+  # Compile-only, on purpose — but the reason changed, so read this before
+  # assuming it still can't run:
   #  - `cargo build` covers the app/library binary.
   #  - `cargo test --lib --no-run` additionally COMPILES the library unit-test
-  #    code on Windows (Windows-clean: the handful of Unix-only cases are
-  #    #[cfg(unix)]-guarded), catching a Windows test-compile break too. `--lib`
-  #    skips the tests/ integration suite (submodule / repo_registry /
-  #    repo_files), which uses unguarded std::os::unix APIs and doesn't compile
-  #    on Windows yet (a separate follow-up).
-  #  - We deliberately do NOT run the tests. The Tauri library's test binary
-  #    transitively links the WebView2/wry native stack and fails to resolve a
-  #    DLL entry point at startup on a stock runner (exit 0xc0000139,
-  #    STATUS_ENTRYPOINT_NOT_FOUND) before any test executes. The cli_shim
-  #    launcher tests are platform-agnostic pure functions already run by the
-  #    Linux `rust` job, so running them here would add nothing regardless.
+  #    code on Windows, catching a Windows test-compile break too.
+  #  - The whole test suite now COMPILES on Windows, and the library test
+  #    binary now STARTS. It used to die at startup with exit 0xc0000139
+  #    (STATUS_ENTRYPOINT_NOT_FOUND) before any test ran, because it imports a
+  #    comctl32 v6 export without the manifest that binds v6; build.rs
+  #    delay-loads comctl32 to fix that, and the two remaining Unix-only cases
+  #    in tests/submodule.rs are #[cfg(unix)]-guarded.
+  #  - We still do NOT run the tests here, and that is now a real choice rather
+  #    than an impossibility. Locally the suite is 613 passed / 88 failed on
+  #    Windows: one library test (a ConPTY round-trip that reads nothing) plus
+  #    an integration cluster concentrated in workdir/merge/rebase/submodule.
+  #    Those are pre-existing Windows gaps that have never been allowed to run
+  #    before, not regressions. Turning the run on is a follow-up that belongs
+  #    with fixing them — flipping it now would just make every PR red.
   # No apt/system-deps step: Windows runners ship the MSVC toolchain and WebView2
   # SDK, and libgit2-sys builds its vendored copy with `cc`, same as on Linux.
   rust-windows:
--- a/src-tauri/build.rs
+++ b/src-tauri/build.rs
@@ -8,5 +8,88 @@ fn main() {
     // reports "Compiling gitcat". Watch the whole icons/ directory so any
     // icon regeneration always re-embeds correctly.
     println!("cargo:rerun-if-changed=icons");
+    delay_load_comctl32_on_windows();
     tauri_build::build()
+}
+
+/// Let the library's unit-test binary START on Windows.
+///
+/// It could not. `cargo test --lib` exited with STATUS_ENTRYPOINT_NOT_FOUND
+/// (0xC0000139) before the harness ran anything, which took `export_bindings`
+/// — the test that regenerates `src/ipc/bindings.ts` — down with it. CI runs
+/// the Rust suite on Linux, so nothing caught it.
+///
+/// Cause: this binary imports `comctl32!TaskDialogIndirect`, which exists only
+/// in comctl32 **version 6**. `C:\Windows\System32\comctl32.dll` is v5 and does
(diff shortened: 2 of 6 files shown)


zangjiucheng (approved):
Approving, and apologies for the wait on this one.

All three fixes are correct, and the interesting one checks out: on git 2.50.1, writing `C:\Temp\GitCat\child` really does land as `url = C:\\Temp\\GitCat\\child` in the file, and `git config --get -f .gitmodules` reads back the single-backslash form. Your argument order parses fine too. The `contains()` to `assert_eq!` change is a real strengthening on Linux and macOS as well, not just a Windows workaround.

Nice use of `let _moved_away_owner` rather than `let _ = ...`. That is exactly where this pattern usually gets written wrong and drops the guard immediately, and the drop order works out since the owner is declared last.

Scope is clean. I went through all six `parent.read(".gitmodules")` sites in `submodule.rs`: the other five compare against literals that escaping cannot affect, and you correctly left them alone. `with_file_name` appears exactly once in the suite, so there was no second instance either.

Two follow-ups, neither blocking:

- The new comment at `submodule.rs:1297-1301` says nothing owned the result, but the test's own `let _ = std::fs::remove_dir_all(&moved_away)` is still there at `:1328`. The success path was always cleaned; the leak only happened when the test panicked earlier, which is what Windows was doing. Worth rewording to that and dropping `:1328` now that `_moved_away_owner` owns it, so the file has a single cleanup story.
- The `cfg(unix)` gate is the right call and loses no coverage today, since the Windows job is compile-only and a tab genuinely cannot exist in a Windows filename. But `validate_pathspec` has no `#[cfg(test)]` module anywhere, so this integration test was the only guard on that regression. A two-line unit test asserting `validate_pathspec("a\tb.txt").is_ok()` would cover it on every platform with no filesystem involved.

You disclosed the stack on #98 at the top of the body, so nothing was hidden. I am merging #98 first and then this, at which point the diff shrinks to the two files it actually owns.

Issue #102, opened 2026-08-21
The ConPTY round-trip test reads nothing from the PTY on Windows


JerryHyun:
`terminal::tests::open_pty_shell_spawns_a_real_shell_and_round_trips_a_command` fails on Windows, and I could not work out why. Filing it with what is known rather than leaving it as an unexplained red test.

Visible only now that the library test binary can start on Windows — see #98.

### What happens

`src/terminal.rs:269`

```
expected echoed output, got: ""
```

The test spawns the platform's default shell through the PTY, writes `echo hello_gitcat_terminal\n`, and waits for that string to come back on the master side. On Windows **nothing at all** arrives — not the echoed command, not the shell's own banner, not a prompt. `collected` is empty, so the reader thread never delivered a single byte before the timeout.

### What has been ruled out

- **Not slowness.** Raising the wait from 5s to 20s changes nothing; the read produces nothing rather than arriving late.
- **Not the harness.** Every other test in the same binary passes (255 of 256). The failure is specific to this one.
- **Not `#98`'s delay-load.** The test never calls into comctl32, and the same failure appears whichever way the binary is linked.

### What has not been tried

- Sending `\r` or `\r\n` instead of `\n`. cmd.exe submits a line on CR, and ConPTY's input translation is not obviously the same as a unix pty's — but an unsubmitted command would still leave the shell's banner in the buffer, and there is no banner either, so this alone does not explain an empty read.
- Reading the master side before writing anything, to see whether the PTY produces output at all on this platform.
- Whether the app's built-in terminal drawer actually works on Windows. If it does, the difference between it and this test is the interesting part; if it does not, this test is reporting a real bug rather than a test-environment problem.

That last one is the question worth answering first — it decides whether this is a test to fix or a feature to fix.

### Note

`open_pty_shell`'s other test (`..._refuses_a_dubious_ownership_path_before_ever_touching_a_pty`) passes on Windows, so the trust gate and the error path are fine; this is specifically about reading PTY output.

---
<sub>Found while making `cargo test` runnable on Windows (#98). Windows 11 / MSVC 14.50; `cargo test -j 2 --lib`.</sub>

Issue #100, opened 2026-08-21
bisect run's good/bad judgement is a shell script, so all four run tests fail on Windows


JerryHyun:
All four `bisect_run_*` tests fail on Windows: the good/bad judgement they hand to `git bisect run` is a shell script, and Windows has no shell to run it.

Visible only now that the Rust suite can start on Windows — see #98.

### Where

`tests/bisect.rs`:

- `bisect_run_converges_via_scripted_good_bad_command` (:180)
- `bisect_run_handles_a_skip_exit_code_and_still_converges` (:228)
- `bisect_run_cancel_stops_the_loop_before_convergence` (:346)
- `bisect_run_start_refuses_a_second_concurrent_call_while_one_is_in_flight`

The first two fail by converging on the wrong commit, which is what a judgement command that cannot run looks like from the outside:

```
assertion `left == right` failed: FIRST-BAD MISMATCH: automated run reported Some("56e1d0c"), expected K 881f5a0
  left:  "56e1d0c"
  right: "881f5a0"
```

The other two fail on timing instead — the run never gets going, so:

```
the test command should have started running before the timeout
```

One of them is timing-sensitive enough to flip between runs, so the suite's total failure count reads 19 on some runs and 20 on others.

### Fix

The tests need a judgement command that exists on the platform they run on. Options, roughly in order of how much they change:

1. Write the judge as a tiny Rust helper binary the test builds and points `bisect run` at. Platform-independent, no shell involved, and the exit codes (0 good / 1 bad / 125 skip) are the whole contract — but it adds a build artifact.
2. Keep a script, and pick the interpreter per platform: `sh -c` on unix, `cmd /c` on Windows. Smaller change, but now there are two scripts to keep in agreement, and the skip-exit-code case has to be expressed twice.
3. `#[cfg(unix)]` the four tests. Honest, and cheap, but it means `bisect run` — the one bisect path with real control flow in it — has no Windows coverage at all.

(1) looks the most durable; (3) is a reasonable holding position if the others are not worth it yet.

### Note

`bisect`'s non-`run` tests all pass on Windows. This is specifically the automated-run path.

---
<sub>Found while making `cargo test` runnable on Windows (#98). Measured with `cargo test -j 2 --no-fail-fast` on Windows 11 / MSVC 14.50.</sub>

PR #104, opened 2026-08-23
One git config for the whole test run, so nested clones get it too

### LINKED ISSUE #99: Nested submodule repos inherit the host's core.autocrlf, failing 8 tests on Windows

JerryHyun:
Closes #99.

**Stacked on #98 and #103** — the diff shows their commits until they merge. Only `d3a1c44` belongs to this one.

Windows: 18 → 9 failing. The nine that remain are #100 (4), #102 (1), and four that need `git-filter-repo` installed — which CI does and my machine did not.

## The problem repo-local config could not reach

`tests/submodule.rs` had ten failures of the same shape: a test compared `"hello\n"` against the `"hello\r\n"` git had just written, because the host runs Git for Windows' default `core.autocrlf=true`.

#98 fixed that class everywhere else by setting `core.autocrlf=false` in each throwaway repo. It cannot work here. `git submodule add` clones the child into an **independent** repository — config is not inherited — and it does so *inside the call under test*:

```rust
let result = block_on(submodule_add(parent.path(), child.path(), "sub".into(), None));
let content = std::fs::read_to_string(parent.dir.join("sub").join("f.txt")).unwrap();
assert_eq!(content, "hello\n");   // reads "hello\r\n"
```

There is no moment between the clone and the read for a test to configure that repo.

## The fix: one config file for the whole run

The fixture now writes its own git config file and points `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` at it **for the process**, so every git in the run reads it — the fixture's, the code under test's, and the ones git itself spawns inside a submodule clone.

That subsumes two things rather than adding a third:

- **The `/dev/null` `git()` used to point at.** Our file blocks the host's `rerere.autoupdate` and friends just as well, and now a git the fixture runs and a git the code under test runs see the *same* global config — rather than one seeing none and the other seeing the host's. That asymmetry is what made this whole class of failure possible.
- **The repo-local `core.autocrlf`/`core.eol` pair #98 added**, now removed as redundant. One mechanism, one explanation, in one place. (This is why the diff touches something the parent PR just introduced — it generalizes it rather than sitting beside it.)

## I checked before turning this knob

Forcing `autocrlf=false` in tests would be the wrong move if the *app* were the thing misbehaving on a CRLF working tree — that is a real thing Windows users would hit, and making the suite green would hide it.

All ten failures are `assertion left == right` on file **content**. Not one is a status or `is_clean()` assertion. The app reported everything correctly; git wrote CRLF because the host is configured to, and the tests read raw bytes. Nothing user-facing is being masked here.

## The sharp edge

`std::env::set_var` in a multi-threaded process. It is mitigated, not eliminated: `OnceLock` makes it happen exactly once, and every test's first act is building a `TempRepo`, so the write lands before that thread spawns any git. A thread already mid-spawn when the very first test initializes could in principle miss it. The reasoning and the escape hatch (hoist it into a pre-main) are in the comment rather than left for someone to rediscover.

## Verified

On Windows: `--test submodule` 49 passed / 0 failed, from 40/9. Full suite 691 passed / 9 failed, up from 682/18.

Nothing changes on Linux, where `autocrlf` is already effectively off — the file just makes that explicit instead of depending on it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,20 +188,24 @@ jobs:
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
   # catches a Windows-only compile break on the PR instead.
   #
-  # Compile-only, on purpose:
+  # Compile-only, on purpose — but the reason changed, so read this before
+  # assuming it still can't run:
   #  - `cargo build` covers the app/library binary.
   #  - `cargo test --lib --no-run` additionally COMPILES the library unit-test
-  #    code on Windows (Windows-clean: the handful of Unix-only cases are
-  #    #[cfg(unix)]-guarded), catching a Windows test-compile break too. `--lib`
-  #    skips the tests/ integration suite (submodule / repo_registry /
-  #    repo_files), which uses unguarded std::os::unix APIs and doesn't compile
-  #    on Windows yet (a separate follow-up).
-  #  - We deliberately do NOT run the tests. The Tauri library's test binary
-  #    transitively links the WebView2/wry native stack and fails to resolve a
-  #    DLL entry point at startup on a stock runner (exit 0xc0000139,
-  #    STATUS_ENTRYPOINT_NOT_FOUND) before any test executes. The cli_shim
-  #    launcher tests are platform-agnostic pure functions already run by the
-  #    Linux `rust` job, so running them here would add nothing regardless.
+  #    code on Windows, catching a Windows test-compile break too.
+  #  - The whole test suite now COMPILES on Windows, and the library test
+  #    binary now STARTS. It used to die at startup with exit 0xc0000139
+  #    (STATUS_ENTRYPOINT_NOT_FOUND) before any test ran, because it imports a
+  #    comctl32 v6 export without the manifest that binds v6; build.rs
+  #    delay-loads comctl32 to fix that, and the two remaining Unix-only cases
+  #    in tests/submodule.rs are #[cfg(unix)]-guarded.
+  #  - We still do NOT run the tests here, and that is now a real choice rather
+  #    than an impossibility. Locally the suite is 613 passed / 88 failed on
+  #    Windows: one library test (a ConPTY round-trip that reads nothing) plus
+  #    an integration cluster concentrated in workdir/merge/rebase/submodule.
+  #    Those are pre-existing Windows gaps that have never been allowed to run
+  #    before, not regressions. Turning the run on is a follow-up that belongs
+  #    with fixing them — flipping it now would just make every PR red.
   # No apt/system-deps step: Windows runners ship the MSVC toolchain and WebView2
   # SDK, and libgit2-sys builds its vendored copy with `cc`, same as on Linux.
   rust-windows:
--- a/src-tauri/build.rs
+++ b/src-tauri/build.rs
@@ -8,5 +8,88 @@ fn main() {
     // reports "Compiling gitcat". Watch the whole icons/ directory so any
     // icon regeneration always re-embeds correctly.
     println!("cargo:rerun-if-changed=icons");
+    delay_load_comctl32_on_windows();
     tauri_build::build()
+}
+
+/// Let the library's unit-test binary START on Windows.
+///
+/// It could not. `cargo test --lib` exited with STATUS_ENTRYPOINT_NOT_FOUND
+/// (0xC0000139) before the harness ran anything, which took `export_bindings`
+/// — the test that regenerates `src/ipc/bindings.ts` — down with it. CI runs
+/// the Rust suite on Linux, so nothing caught it.
+///
+/// Cause: this binary imports `comctl32!TaskDialogIndirect`, which exists only
+/// in comctl32 **version 6**. `C:\Windows\System32\comctl32.dll` is v5 and does
+/// not export it; v6 lives in a side-by-side assembly the loader binds only for
+/// a process whose manifest declares a `Microsoft.Windows.Common-Controls`
+/// 6.0.0.0 dependency. `tauri_build` embeds that manifest in the APP binary and
+/// nothing embeds it in a test binary.
+///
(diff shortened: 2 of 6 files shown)


zangjiucheng (changes requested):
The mechanism is right, and it is the only lever that reaches `git submodule add`'s nested clone, so no argument from me there. Worth saying out loud as well: this quietly fixes a real side effect, since `trust.rs:38-39` runs `git config --global --add safe.directory` through the inherited env and was writing into the developer's actual `~/.gitconfig` during test runs.

Two things to fix, both test-only.

`tests/git_config.rs:232` ends with `remove_var`. That was correct on dev, where the ambient value is unset, but this PR makes the ambient value the run's own config file, so the same line now destroys the invariant instead of restoring it. `global_git_config()` is a `OnceLock`, so nothing ever sets it back and every later test in that binary falls through to the real `~/.gitconfig`.

```rust
// tests/git_config.rs — restore, don't remove (needs global_git_config() to be pub)
unsafe { std::env::set_var("GIT_CONFIG_GLOBAL", common::global_git_config()); }
```

`tests/identity.rs`'s `real_global()` doc says it reads the ambient value "precisely so we can prove it's untouched", and after this change that ambient value is the fixture. So `get_git_identity_falls_back_to_the_real_global_identity_when_local_is_unset` and the partial-override test compare `None` against `None` on every machine, and `set_git_identity_writes_local_config_and_never_touches_the_real_global_config` can no longer detect a write escaping to global.

The logic itself stays covered either way, since `resolve_falls_back_to_global_when_local_is_entirely_unset` in `src/identity.rs` is a pure unit test. What is lost is the end-to-end evidence that git really reads `--global`. Either give the run's config a distinguishable identity:

```rust
// common/mod.rs, global_git_config() — alongside the [core] block
"[user]\n\tname = GitCat Global\n\temail = global@gitcat.example\n"
```

which brings those two tests back to life, or decide the trade is not worth it (it would mask the empty-GECOS case `apply_test_config` sets a local identity for) and rewrite the two doc comments to the new contract instead. Either is fine. What I do not want is the docs still describing the old one.

Two nits: the config file is the one path in this file not using the pid+nanos+SEQ naming, and `fs::write(...).expect(...)` turns a collision into a panic that takes the whole binary down at the first `TempRepo`. And the `set_var` comment names the wrong hazard, the danger is a concurrent `getenv`, and `env::temp_dir()` a few lines up is one.

One question I would like answered in the doc comment either way: `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` are git-CLI variables and libgit2 does not honour them, it resolves global config from `$HOME`/`%USERPROFILE%` directly. The backend calls `repo.statuses()` through git2 in `workdir.rs:474,1036`, `dashboard.rs:132` and `submodule.rs:1264,1384`. Did you confirm the git2 side also sees `autocrlf=false` on Windows, or is the suite green because the CLI writes LF working trees and libgit2's clean filter is a no-op on LF regardless? Right now "one git config for the whole run" is true for the CLI only.

PR #105, opened 2026-08-24
Make the bisect run-loop test portable; gate the three that cannot be

JerryHyun:
Partially addresses #100 — **deliberately not closing it**, see below.

**Stacked on #98, #103 and #104** — the diff shows their commits until they merge. Only `6c3f071` belongs to this one.

Windows: 9 → 5 failing. Of the five left, four need `git-filter-repo` installed (CI does; my machine did not) and one is #102.

## The code under test was never the problem

`run_test_command` already shells out through `cmd /C` on Windows and `sh -c` elsewhere, and documents its own exit-code caveat there. What could not run on Windows was the **scaffolding** — the judge commands these tests hand to `git bisect run` are POSIX shell.

## The convergence test is now portable

It asked `test ! -f bug.txt`, a builtin cmd.exe does not have. It now asks the same question of a program both shells can run:

```
git cat-file -e HEAD:bug.txt && exit 1 || exit 0
```

`&&`, `||` and `exit` mean the same thing in both shells, and during a bisect the working tree matches HEAD, so asking HEAD is equivalent to asking the filesystem. Verified under `cmd /C` directly — absent → `0` (good), present → `1` (bad).

This is the valuable one of the four: it is what actually drives `run_bisect`'s loop to convergence, and it now does so on Windows.

## Three are `#[cfg(unix)]`, and that is a coverage loss worth naming

Skip handling, cancellation and the concurrency refusal each need two things from their shell:

- **state that survives between separate invocations** (a marker file), and
- **a bounded block**, so the main thread can observe the run mid-flight.

There is no spelling of those that both `sh -c` and `cmd /C` accept. cmd has no `sleep`, and `timeout /t` wants a console this code deliberately withholds (`no_console_window`). Writing two scripts per test and keeping them in agreement would buy coverage of the scaffolding, not of the loop.

So: the run loop is now exercised on Windows; **skip, cancel and the concurrency refusal are not**. The comment on the gated tests says exactly that, and points at the issue.

## Why the issue stays open

The remaining answer is a real judge *executable* rather than a shell string — a small program with the exit-code contract baked in, which both shells invoke identically. That is a larger change than this one (it needs a build target and a way for an integration test to find it), and it did not belong bundled in here. Closing #100 on a partial fix would misreport what shipped.

Imports used only by the gated tests are gated with them, so Windows builds warning-free.

## Verified

On Windows: `--test bisect` 5 passed / 0 failed (was 4 passed / 4 failed), no compile warnings. Full suite 692 passed / 5 failed, up from 691/9.

Nothing changes on unix: the three gated tests still run there, and the portable judge is equivalent to the one it replaced.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,20 +188,24 @@ jobs:
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
   # catches a Windows-only compile break on the PR instead.
   #
-  # Compile-only, on purpose:
+  # Compile-only, on purpose — but the reason changed, so read this before
+  # assuming it still can't run:
   #  - `cargo build` covers the app/library binary.
   #  - `cargo test --lib --no-run` additionally COMPILES the library unit-test
-  #    code on Windows (Windows-clean: the handful of Unix-only cases are
-  #    #[cfg(unix)]-guarded), catching a Windows test-compile break too. `--lib`
-  #    skips the tests/ integration suite (submodule / repo_registry /
-  #    repo_files), which uses unguarded std::os::unix APIs and doesn't compile
-  #    on Windows yet (a separate follow-up).
-  #  - We deliberately do NOT run the tests. The Tauri library's test binary
-  #    transitively links the WebView2/wry native stack and fails to resolve a
-  #    DLL entry point at startup on a stock runner (exit 0xc0000139,
-  #    STATUS_ENTRYPOINT_NOT_FOUND) before any test executes. The cli_shim
-  #    launcher tests are platform-agnostic pure functions already run by the
-  #    Linux `rust` job, so running them here would add nothing regardless.
+  #    code on Windows, catching a Windows test-compile break too.
+  #  - The whole test suite now COMPILES on Windows, and the library test
+  #    binary now STARTS. It used to die at startup with exit 0xc0000139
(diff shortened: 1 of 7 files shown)


zangjiucheng:
This is clean and the scope is exactly what the description says, but #108 removes the reason it exists: with `raw_arg`, the three `cfg(unix)` gates are not needed, so landing this on its own ships a gate your own later PR proves unnecessary. Could you fold 6c3f071 into #108 and close this one?

To be clear about what the gating was not, since it reads worse than it is: those three tests have never run on Windows. CI is compile-only there and before #98 the integration targets did not even build, so nothing is being switched off.

The portable judge is worth keeping. One tweak, since the current form prints `fatal: path 'bug.txt' does not exist in 'HEAD'` to stderr on every good probe and a passing run ends up full of fatal-looking lines:

```rust
// same contract, same two-shell portability, no fatal noise
"git rev-parse -q --verify HEAD:bug.txt && exit 1 || exit 0"
```

Also the 13-line rationale comment is repeated verbatim above all three gated tests; one copy plus a back-reference would read better.

Separately, thank you for the repro in #100. It failing as a first-bad mismatch rather than an abort is direct evidence that `classify_exit`'s documented Windows gap is real and silently converges on the wrong commit, so I have filed that as its own issue.

PR #106, opened 2026-08-24
Answer ConPTY's cursor-position query, as a terminal must

### LINKED ISSUE #102: The ConPTY round-trip test reads nothing from the PTY on Windows

JerryHyun:
Closes #102.

**Stacked on #98, #103, #104 and #105.** With this one in, the Rust suite is **697 passed / 0 failed on Windows** — from a suite that could not start at all.

## It is explained, and the app was never broken

The test read *nothing* from the PTY: not the echoed command, not cmd.exe's banner, not a prompt. Raising the wait to 20s changed nothing, which is what made it look mysterious.

ConPTY opens by asking the terminal where its cursor is — a Device Status Report, `ESC[6n` — and **blocks until something answers**. Reproduced outside the app with a standalone `portable-pty` program driven exactly like `open_pty_shell`:

| | bytes read | contents |
|---|---|---|
| no reply (what the test did) | **4** | the `ESC[6n` itself, and nothing further |
| replying `ESC[1;1R` first | **292** | banner, prompt, the echoed command, its output |

In the app, xterm.js answers that query automatically — which is exactly why the terminal drawer works on Windows. This test *is* the terminal, and answering is part of the job.

That also answers the question the issue said should be settled first ("does the built-in terminal actually work on Windows?"): yes, and no GUI run was needed to establish it — the mechanism accounts for both behaviours.

## Windows-only, deliberately

No unix pty asks this. And there the escape would not be consumed by anything: it would land in the line buffer and be prefixed to the command. That would *still* satisfy the assertion, because the shell's "not found" message quotes the command back — green while testing nothing. Worse than red.

Also switched the write to CRLF: cmd.exe submits a line on CR, and a unix pty translates CR to NL on input (ICRNL), so it reads the same on both.

## On the `filter_repo` four

Earlier PRs in this stack reported four failures as "not a repo problem — `git-filter-repo` is not installed locally, and CI installs it with pipx". Confirmed: installing it turned all four green with no code change. That is why the total here is 0 rather than 4.

## Verified

On Windows: `--lib` 256 passed / 0 failed; whole suite `cargo test -j 2 --no-fail-fast` 697 passed / 0 failed. Unaffected on unix — the reply is `#[cfg(windows)]`, and CR-submitted lines behave identically there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,20 +188,24 @@ jobs:
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
   # catches a Windows-only compile break on the PR instead.
   #
-  # Compile-only, on purpose:
+  # Compile-only, on purpose — but the reason changed, so read this before
+  # assuming it still can't run:
   #  - `cargo build` covers the app/library binary.
   #  - `cargo test --lib --no-run` additionally COMPILES the library unit-test
-  #    code on Windows (Windows-clean: the handful of Unix-only cases are
-  #    #[cfg(unix)]-guarded), catching a Windows test-compile break too. `--lib`
-  #    skips the tests/ integration suite (submodule / repo_registry /
-  #    repo_files), which uses unguarded std::os::unix APIs and doesn't compile
-  #    on Windows yet (a separate follow-up).
-  #  - We deliberately do NOT run the tests. The Tauri library's test binary
-  #    transitively links the WebView2/wry native stack and fails to resolve a
-  #    DLL entry point at startup on a stock runner (exit 0xc0000139,
-  #    STATUS_ENTRYPOINT_NOT_FOUND) before any test executes. The cli_shim
-  #    launcher tests are platform-agnostic pure functions already run by the
-  #    Linux `rust` job, so running them here would add nothing regardless.
+  #    code on Windows, catching a Windows test-compile break too.
+  #  - The whole test suite now COMPILES on Windows, and the library test
+  #    binary now STARTS. It used to die at startup with exit 0xc0000139
+  #    (STATUS_ENTRYPOINT_NOT_FOUND) before any test ran, because it imports a
+  #    comctl32 v6 export without the manifest that binds v6; build.rs
+  #    delay-loads comctl32 to fix that, and the two remaining Unix-only cases
+  #    in tests/submodule.rs are #[cfg(unix)]-guarded.
+  #  - We still do NOT run the tests here, and that is now a real choice rather
+  #    than an impossibility. Locally the suite is 613 passed / 88 failed on
+  #    Windows: one library test (a ConPTY round-trip that reads nothing) plus
+  #    an integration cluster concentrated in workdir/merge/rebase/submodule.
+  #    Those are pre-existing Windows gaps that have never been allowed to run
+  #    before, not regressions. Turning the run on is a follow-up that belongs
+  #    with fixing them — flipping it now would just make every PR red.
   # No apt/system-deps step: Windows runners ship the MSVC toolchain and WebView2
   # SDK, and libgit2-sys builds its vendored copy with `cc`, same as on Linux.
   rust-windows:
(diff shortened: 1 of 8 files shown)


zangjiucheng (approved):
Approving. Sorry for the slow turnaround.

Good piece of debugging, and the fix is the right shape. Playing terminal well enough to answer ConPTY's startup DSR is precisely what the test was missing, and the 4-bytes-versus-292-bytes measurement makes it checkable rather than merely plausible. The cross-check against the app holds up from our side too: `Terminal.svelte:52` pipes `xterm.onData` into `terminalWrite` and xterm.js answers `CSI 6n` on its own, so "the drawer works because xterm.js replies" is verifiable without a GUI run.

Keeping the reply `#[cfg(windows)]` was the right instinct, and for the reason you gave: on a unix pty the escape would be quoted back by the shell's not-found message and the assertion would pass while testing nothing.

Follow-ups, none blocking:

- Take that instinct one step further. `collected.contains("hello_gitcat_terminal")` at `terminal.rs:290` is satisfied by the terminal's echo of the typed line on both platforms, so it cannot distinguish "the shell ran it" from "the bytes came back". Requiring the marker twice would close that, and you have the Windows box to confirm cmd.exe gives two.
- Please fix the `ci.yml` comment. Lines 202-208 still say the Windows suite is 613 passed / 88 failed and name a ConPTY round-trip that reads nothing as one of the failures. That is this test, and this commit fixes it. Shipping a rationale your own commit contradicts costs the next person an afternoon.
- The bigger one, related: nothing in CI runs any of this. The Windows job stops at `cargo test --locked --lib --no-run`, so the `#[cfg(windows)]` reply is compiled and never executed. If `--lib` really is 256/0, flipping that step to a real run gives the fix a regression guard. That is the whole point of the job.

On ordering: since #98, #103, #104 and #105 are all still open against dev, merging this merges the lot. I would rather land them in order and approve #98's `/DELAYLOAD` trade as #98. Also worth rebasing, since dev has rewritten `parse_first_bad` in `git_bisect.rs` since you branched.

PR #108, opened 2026-08-24
Pass the bisect test command to cmd.exe unmangled; un-gate the run tests

### LINKED ISSUE #107: Automated bisect marks every commit bad on Windows when the test command is quoted
On Windows, an automated bisect whose test command is quoted reports every commit as **bad** — the command never runs, and `git bisect run` converges on the wrong commit with no error shown.

Quoting is not exotic here: any path with a space needs it. `"C:\Program Files\ci\test.bat"`, or a checkout under `C:\Users\First Last\…`.

### Where

`src-tauri/src/git_bisect.rs`, `run_test_command`:

```rust
let mut c = Command::new("cmd");
c.arg("/C").arg(command);
```

Rust escapes a `"` inside an argument as `\"` — the MSVC convention. cmd.exe does not know it, and reads the backslash as part of the program name. Measured with a command whose path contains a space:

| how the string is passed | cmd's response | exit |
|---|---|---|
| `.arg(command)` | `'\"C:\…\judge.bat\"' is not recognized as an internal or external command` | **1** |
| `.raw_arg(command)` | runs | **0** |

### Why it is worse than a failed step

`classify_exit` treats any code in `1..=127` as **bad**. So exit 1 — "I could not run your command at all" — is indistinguishable from "this commit is broken". The bisect does not abort, does not warn; it just marks the commit bad and keeps going. The user gets a confident, wrong first-bad commit.

`classify_exit`'s own doc comment already anticipated this shape:

> other Windows failure modes that return a code in 1..=127 are NOT distinguished from a genuine "bad" result here and will be misclassified as "bad" rather than "abort"

This is a concrete instance of it, and one that is fixable at the source rather than by guessing at exit codes.

### Fix

Use `std::os::windows::process::CommandExt::raw_arg` for the command string on Windows, so it reaches cmd.exe exactly as the user wrote it — the same text they would type at a prompt. The unix `sh -c` branch is unaffected: `arg` is correct there.

### Note

Found while making the bisect tests run on Windows (#100): the tests could not invoke a judge program by absolute path for exactly this reason. The test suite hitting it first is luck — a user with a space in their path hits it with no test involved.

---
<sub>Measured on Windows 11 / MSVC 14.50 with a standalone probe comparing `.arg` and `.raw_arg` against a batch file under a path containing a space.</sub>

JerryHyun:
Closes #107, and finishes #100.

**Stacked on #98, #103, #104, #105 and #106.** Only `9e5971d` belongs to this one. With it in, the Rust suite is **700 passed / 0 failed on Windows**.

## The app bug

`run_test_command` built its Windows invocation with `Command::arg`, which escapes an embedded `"` as `\"` — the MSVC convention. cmd.exe does not share it, and reads the backslash as part of the program name.

So any **quoted** command — meaning any path with a space, `"C:\Program Files\ci\test.bat"`, or a checkout under `C:\Users\First Last\…` — reached cmd as `\"C:\Program Files\…\"` and came back:

```
'\"C:\Program Files\ci\test.bat\"' is not recognized as an internal or external command
```

exit **1**.

And exit 1 sits inside `classify_exit`'s `1..=127` arm, which means **bad**. A command that never ran condemned the commit — silently, with no abort and no warning — and the bisect converged on the wrong first-bad commit. Measured both ways against a batch file under a path with a space:

| | exit |
|---|---|
| `.arg(command)` | 1 (misread as "bad") |
| `.raw_arg(command)` | the command's own status |

`classify_exit`'s doc comment had already predicted this *shape* of failure ("other Windows failure modes that return a code in 1..=127 are NOT distinguished from a genuine bad result"). It now records that this instance was found and closed at the source, while the general gap remains.

The unix `sh -c` branch is untouched — `arg` is correct there.

## Finishing #100

With that fixed, the three automated-run tests #105 had to gate can run on Windows after all.

They need two things from their judgement that no shell provides on both platforms: state that survives between invocations, and a bounded block so the test's main thread can observe a run mid-flight. So the judgement is now a program — `examples/bisect_judge.rs` — which `sh -c` and `cmd /C` invoke identically.

An **example** rather than a `[[bin]]`, so it never ships in the app. `cargo test` builds examples, and the test finds it next to its own binary: `target/<profile>/examples/` is a sibling of the `deps/` the test binary lives in, which stays correct under `CARGO_TARGET_DIR`, `--release` and cross-compilation. Its arguments are deliberately free of spaces and metacharacters, since quoting is the other thing the two shells disagree about.

This is why the diff re-touches what #105 just gated: #105 was the honest answer while the app bug was still there; this is the better one now that it is not.

## Also: LNK4199

#98's `/DELAYLOAD:comctl32.dll` is unscoped, because the target that needs it cannot be addressed on its own. It therefore also reaches targets with no comctl32 imports — and this new example, which links no part of the app, is the first such target, so the linker started warning that the flag was ignored. The flag being a no-op there is the intended outcome, so `/IGNORE:4199` goes alongside it, with the reason.

## Verified

On Windows: `--test bisect` 8 passed / 0 failed, including all three previously-gated tests. Whole suite `cargo test -j 2 --no-fail-fast` **700 passed / 0 failed**, no warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,20 +188,24 @@ jobs:
   # otherwise unverified until a tagged release runs the 6-platform matrix. This
   # catches a Windows-only compile break on the PR instead.
   #
-  # Compile-only, on purpose:
+  # Compile-only, on purpose — but the reason changed, so read this before
+  # assuming it still can't run:
   #  - `cargo build` covers the app/library binary.
   #  - `cargo test --lib --no-run` additionally COMPILES the library unit-test
-  #    code on Windows (Windows-clean: the handful of Unix-only cases are
-  #    #[cfg(unix)]-guarded), catching a Windows test-compile break too. `--lib`
-  #    skips the tests/ integration suite (submodule / repo_registry /
-  #    repo_files), which uses unguarded std::os::unix APIs and doesn't compile
-  #    on Windows yet (a separate follow-up).
-  #  - We deliberately do NOT run the tests. The Tauri library's test binary
-  #    transitively links the WebView2/wry native stack and fails to resolve a
-  #    DLL entry point at startup on a stock runner (exit 0xc0000139,
-  #    STATUS_ENTRYPOINT_NOT_FOUND) before any test executes. The cli_shim
-  #    launcher tests are platform-agnostic pure functions already run by the
-  #    Linux `rust` job, so running them here would add nothing regardless.
+  #    code on Windows, catching a Windows test-compile break too.
+  #  - The whole test suite now COMPILES on Windows, and the library test
+  #    binary now STARTS. It used to die at startup with exit 0xc0000139
+  #    (STATUS_ENTRYPOINT_NOT_FOUND) before any test ran, because it imports a
+  #    comctl32 v6 export without the manifest that binds v6; build.rs
+  #    delay-loads comctl32 to fix that, and the two remaining Unix-only cases
+  #    in tests/submodule.rs are #[cfg(unix)]-guarded.
+  #  - We still do NOT run the tests here, and that is now a real choice rather
+  #    than an impossibility. Locally the suite is 613 passed / 88 failed on
+  #    Windows: one library test (a ConPTY round-trip that reads nothing) plus
+  #    an integration cluster concentrated in workdir/merge/rebase/submodule.
+  #    Those are pre-existing Windows gaps that have never been allowed to run
+  #    before, not regressions. Turning the run on is a follow-up that belongs
+  #    with fixing them — flipping it now would just make every PR red.
   # No apt/system-deps step: Windows runners ship the MSVC toolchain and WebView2
   # SDK, and libgit2-sys builds its vendored copy with `cc`, same as on Linux.
   rust-windows:
--- a/src-tauri/build.rs
+++ b/src-tauri/build.rs
@@ -92,4 +92,11 @@ fn delay_load_comctl32_on_windows() {
     // Provides __delayLoadHelper2, which the delay-load stubs call. Without it
     // the link fails with LNK2001 on that symbol.
     println!("cargo:rustc-link-arg=delayimp.lib");
+    // LNK4199: "/DELAYLOAD:comctl32.dll ignored; no imports found". These flags
+    // are unscoped because the target that needs them — the lib's own unittest
+    // binary — cannot be addressed on its own (see above), so they also reach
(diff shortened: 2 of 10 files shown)


zangjiucheng (approved):
Approving, and sorry for leaving this sitting.

The core fix is right and the diagnosis is convincing. `Command::arg` applying MSVC argv quoting to a string that cmd.exe then re-parses under completely different rules, landing in `classify_exit`'s `1..=127` BAD arm, is exactly the kind of bug that produces a confident wrong answer, and fixing it at the source with `raw_arg` rather than guessing at exit codes is the right call. The resulting quoting matches cmd's documented `/C` rules in both directions: the space-containing path (rule 1, quotes preserved) and the space-free path (rule 2, pair stripped) both resolve correctly, and the unix branch is genuinely untouched.

This is the one in the stack I care most about, because a bisect test command that never ran was being reported BAD, which converges on the wrong first-bad commit. That is a shipped user-visible bug, not test hygiene.

Follow-ups:

- `tests/bisect.rs:465` still runs `test ! -f bug.txt` in `bisect_run_start_refuses_a_second_concurrent_call_while_one_is_in_flight`, which this PR un-gates for Windows. That is the same POSIX-only construct your own comment at line 205 flags as unavailable in cmd.exe. It passed for you because Git for Windows' `usr/bin` was on PATH; on a stock box cmd returns 9009, `classify_exit` maps that to Abort, and the assert at `:468` fails. The portable form you already use at line 209 fixes it. Nothing turns red in CI since the Windows job stays compile-only, but it does mean the 700/0 figure is environment-dependent.
- Ergonomics: `cargo test --test bisect` does not build examples, so that common invocation trips the assert on a clean tree. Worth naming the remedy in the message. The `examples/` choice over a `[[bin]]` is well reasoned, and `judge_exe()`'s `current_exe`-relative resolution is more robust than the usual target-dir guessing.
- Worth filing as a follow-up: `tool_settings.rs:764` and `plugin_exec.rs:417` build the identical `cmd /C` plus `.arg` invocation and have the identical bug. If that becomes a shared helper, `plugin_exec`'s `windows_cmd_unsafe` fail-closed guard has to survive it.
- Since this is the commit that makes the Windows suite green, the `ci.yml` comment at line 203 saying 613 passed / 88 failed is now stale.

Merge order: this is the tip of the #98/#103/#104/#105/#106 stack, so merging it alone brings all of them; only 9e5971d is this PR's own work. I would also fold #105 into this one, since its three `cfg(unix)` gates exist only because of the bug you fix here.

PR #109, opened 2026-08-24
Terminal opens in C:\Windows instead of the repository

### LINKED ISSUE #42: [Bug]: Terminal not open in correct folder with windows version
### GitCat version

nightly build

### Operating system

Windows

### Architecture

Apple Silicon (arm64)

### OS version

_No response_

### Steps to reproduce

Image

### What you expected

Should open terminal in correct path with windows and wsl env. 

### What actually happened

Open with the path start with "\\?" which is incorrect.

### Logs (optional)

```shell

```

JerryHyun:
Closes #42.

**Stacked on #98, #103, #104, #105, #106 and #108** — the diff shows their commits until they merge. Only `4eb2a46` and `2292a0c` belong to this one.

## What was wrong

Open a repository from the Dashboard, then open the built-in terminal. It starts in `C:\Windows`, not in your repository. Every command you type runs in the wrong place.

The screenshot in the issue has cmd.exe explaining itself:

```
'\\?\C:\Users\...'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
```

## Why

Windows has a second way to write a path, starting with `\\?\`. It exists so programs can use very long paths, and the Windows file functions accept it — but almost nothing else does. cmd.exe reads it as a network path, refuses to start there, and quietly falls back to `C:\Windows`.

GitCat was storing paths in that form. `repo_registry::normalize` asks Windows to resolve each tracked repository's path to its full real form, and on Windows that answer always comes back as `\\?\C:\...`. So every repository opened from the Dashboard — the only place a path comes from the registry — handed the terminal a folder the shell would not accept.

The app already had a helper for exactly this, `windows::strip_windows_verbatim_prefix`, added when `gitcat .` hit the same thing from the command line. `normalize` just never called it. Now it does.

`wsl.rs` had even worked around half of this already: its comment explains that `normalize` produces `\\?\UNC\...` for a WSL repo, and it strips that prefix before deciding whether a path is a WSL one. The `\\?\C:\` half had nobody looking after it.

## Existing users

`load_from` now rewrites old entries as it reads them. Without that, a registry written by an earlier build would stop matching the key `normalize` returns now: every repository would be tracked a second time on its next open, and each one's saved branch-visibility filter would read as "never set".

It also merges the two spellings of one repository onto a single entry — a path that failed to resolve was already stored in the plain form, so both could exist side by side — keeping whichever was opened more recently.

## The WSL half

Stripping the prefix turns a WSL repository's stored path from `\\?\UNC\wsl.localhost\...` into `\\wsl.localhost\...`. That is still a network path, and cmd.exe refuses it for the same reason. Measured directly, with a real network path:

```
working directory = \\localhost\C$\Temp
-> the process reports C:\Windows
-> "UNC paths are not supported. Defaulting to Windows directory."
```

So a WSL repository now opens the distribution's own shell, already in the repository, with `wsl.exe -d <distro> --cd <linux path>`. That is the same decision `wsl::git_command` already makes for every git call GitCat runs against a WSL repository — this just applies it one level up, to the terminal.

It is also the better answer even setting the folder problem aside. A Windows shell sitting on a WSL share gets you the Windows `git` reaching across into the distribution's filesystem, which is precisely what `wsl::git_command` exists to avoid.

A real network share (`\\server\share\...`) deliberately stays on the ordinary path. It is not WSL, and handing it to `wsl.exe` would be nonsense. cmd.exe still cannot use it as a working directory — that is a separate gap, left visible rather than papered over.

## What I checked, and how

| | |
|---|---|
| The five `pty_command_for` tests | run, 8/8 in `--lib terminal::` |
| The three registry migration tests | run, 9/9 in `--lib repo_registry::` |
| cmd.exe really does refuse a network path | measured — a process started at `\\localhost\C$\Temp` reports `C:\Windows` |
| **WSL routing, end to end** | **run against a real distribution** |
| The whole Rust suite | 707 passed, 0 failed |
| The terminal in the running app | **opens in the repository** — checked by hand on Windows |

The WSL check was not a unit test. I made a git repository inside Ubuntu, confirmed Windows sees it at `\\wsl.localhost\Ubuntu\home\...`, parsed that path the way the app does, built the same command, and spawned it through a PTY. The shell came up inside the repository:

```
cwd=/home/.../gitcat-wsl-probe
hsh858@...:~/gitcat-wsl-probe$
git rev-parse --is-inside-work-tree -> true
```

One caveat on the suite: `tests/file_history.rs`'s slowest test failed once with "failed to rename lockfile" and passed on a re-run. That is an intermittent Windows file-locking failure that has nothing to do with this change — mentioning it so a red run is not mistaken for one.

## One thing I noticed and did not fix

If the WSL distribution is asleep, the first `wsl.exe` call can fail with `Wsl/Service/E_UNEXPECTED` while it starts up; the next one succeeds. It cost me a while, because a call without `-d` succeeded at the same moment and made it look like the distribution name was wrong.

That means a user opening a terminal on a WSL repository may see the first attempt fail. `open_pty_shell` currently reports the spawn error as-is. Whether that deserves a retry or a friendlier message is its own decision, so it is not in here.


zangjiucheng (approved):
Approving, and apologies for the delay in getting to this.

This is aimed at the right layer. `repo_registry::normalize` really was the odd one out, since `instance_focus::canonical_key` and `windows::classify_initial_arg` both already did canonicalize-then-`strip_windows_verbatim_prefix`. Making `normalize` the third is the consistent fix rather than a patch over the symptom. Including the `load_from` migration is the part I would have expected to be missed, and putting it there rather than at each call site is right: that is the single choke point every registry command reads through, and it is idempotent, so it needs no schema bump.

Follow-ups:

- `migrate_verbatim_paths` (`repo_registry.rs:190`): the collapse does `*kept = repo`, which throws away the losing row's `visible_local_branches`, `visible_remote_branches`, `auto_branch_visibility` and `repo_summary_shown`. That case is more reachable than it looks for WSL users, because `normalize` falls back to the raw string when `canonicalize` fails, which is exactly what happens when the distro is stopped. So a WSL repo can genuinely have a plain row carrying the real settings and a verbatim row that is merely more recent. Taking the max timestamp but coalescing the other fields, plus a collapse case where the older row holds the settings, would close it.
- Consistency nit in `terminal.rs`: `open_pty_shell` runs `trust::open_repo` on the raw path while `pty_command_for` strips its own copy, so the gate and the spawn look at two different strings. Strip once at the top and pass the result to both.
- Question on `--cd`: it is a Store-WSL-era flag, so on an older inbox WSL on Windows 10 `wsl.exe` prints usage and exits rather than opening a shell. Neither old nor new behaviour works there so it is not a regression, but a fallback would be cheap insurance.

Worth noting for the changelog: since every Dashboard-opened path came out of `normalize`, this also fixes the cmd.exe-refuses-the-cwd problem in `git_bisect.rs:773`'s bisect-run command for local Windows repos. WSL repos still hit it there, which is probably its own issue.

One request on packaging: only 4eb2a46 and 2292a0c belong to this PR, the rest is the #98 through #108 stack. Since this closes a shipped user-visible bug and is subject-matter independent of the test work, I would rather you rebase those two commits onto dev as a standalone PR than have it wait behind five test-hygiene PRs. Say the word and I will merge it early.

PR #110, opened 2026-08-24
feat: right-click menus for file rows, folder rows and the open repository

JerryHyun:
Right-click a file, a folder, or the repo name in the topbar, and get a menu.

Rebased onto `dev`, so the diff is this branch's own commits rather than the stack it used to carry.

## Demo

https://github.com/user-attachments/assets/0bd63156-3b9d-468c-bccb-9d37e4762797

Recorded on Windows: the menus on a file row, a folder row and the repo chip, and where each of the file-manager actions actually lands.

## What you get

| Where | Items |
|---|---|
| A commit's changed file | Blame · History · External diff · — · Show in \<file manager\> · Copy path · Copy full path |
| A commit's folder | Open in \<file manager\> · Copy path · Copy full path |
| A working-tree file, unstaged | Stage · Discard · — · Show in \<file manager\> · Copy path · Copy full path |
| A working-tree file, staged | Unstage · — · Show in \<file manager\> · Copy path · Copy full path |
| A working-tree folder | Stage/Unstage folder · — · Open in \<file manager\> · Copy path · Copy full path |
| The topbar repo chip | Terminal · Open in \<file manager\> · Copy path |

Each surface's menu carries the actions that surface already shows as icon buttons, then the path items. A menu offering *less* than the visible buttons is a strange thing to discover. Working-tree folders get no Discard — the button row has none either, and a typed-confirm scrim naming a whole directory is a much larger decision than discarding one file. Neither do staged rows: a staged row's discard would act on the worktree, which either fails outright or throws away the unstaged edits and leaves the staged change alone.

## One menu, not a fourth one

Workdir already had a hand-rolled right-click popover, and Sidebar a second. Two more surfaces would have meant four places to fix the next positioning or dismissal bug in, so this adds one shared island (`src/islands/contextmenu/`) that owns the backdrop, Escape/scroll/resize dismissal, and closing before an item runs — several of these actions open a dialog whose first click the backdrop would otherwise swallow.

Workdir's popover is migrated onto it and its markup deleted. **Sidebar's menus are deliberately left alone** — they are unrelated to this change, and rewriting working code for tidiness is how you get a regression you cannot explain.

The shared menu also flips away from the viewport edge rather than clamping. The commit file list sits at the bottom of the window; a six-item menu opening downward from there ran off the fold, and with the backdrop swallowing scroll those items were simply unreachable.

## Why reveal and open go through Rust

`@tauri-apps/plugin-opener` ships JS bindings for both, and `revealItemInDir` is already allowed by this app's `opener:default` capability. The **open** half is the problem: it needs `allow-open-path`, whose scope check (`scope.rs`'s `is_path_allowed`) ends in `self.allowed.iter().any(...)` — false for an empty allow list. So the capability would have to carry one, and the only list broad enough for "whichever repo the user has open" is `**`.

That grants the webview the right to ask the OS to open **any** path with its default application, and this app runs third-party plugin code in that webview (`plugin_exec.rs`). Both actions live in `file_manager.rs` instead, behind `trust::open_repo` — the same gate `terminal.rs` uses for the considerably more powerful shell it spawns.

The commands take the repo path and the file's repo-relative path separately, and join them server-side with a containment check. Callers pass paths from this app's own file lists, so a traversal would be a bug rather than an attack — but the check is two lines and the failure it prevents is the kind that is embarrassing to explain afterwards.

## The one platform branch

`legacy/platform.ts` is this app's first place where copy varies by OS, and is meant to stay the only one. Everything else is written once — the keyboard hints say ⌘ on Windows too. That is a defensible house style for a shortcut glyph, but not for a proper noun: telling a Mac user to look in "File Explorer" names an application that is not on their machine.

Sniffed from the user agent rather than asked of the backend, so it stays synchronous (menu labels render during a right-click, with no await to hang them on) and works in browser design mode where there is no backend at all. Anything unrecognized gets "file manager", which is true on any desktop.

**Two verbs, deliberately.** A file is *revealed* — its folder opens with the file selected. A folder is *opened* — you land inside it. The backend calls differ to match (`reveal_item_in_dir` vs `open_path`), so "Show in Finder" on a folder would be a lie about where it takes you.

That is why folder rows get their own item builder rather than reusing the file one, and why `open_repo_in_file_manager(repo)` is generalized here to `open_dir_in_file_manager(repo, relative)` — an empty `relative` means the repo itself. The topbar chip and a folder row then go through one command instead of two that would have to be kept in step.

One rendering rule came with it: `ContextMenu.svelte` now drops a `separatorBefore` on the *first* item. The shared builders mark their group break on their own first entry, which is right when a surface stacks its actions above them and a stray top border when it does not — exactly a commit tree's folder rows, where the path items are the whole menu.

## One thing found while wiring this up

The two plugin calls disagree about the Windows extended-length `\\?\C:\…` form. `reveal_item_in_dir` runs its argument through `dunce::simplified`, which drops the prefix; `open_path` only stats the path — which succeeds, that being exactly what the prefix is for — and then hands it to `ShellExecute`, which does not understand it. Same input, one works and one does not.

Nothing upstream produces that form any more: `repo_registry::normalize` and the `gitcat .` argument classifier both strip it. So `file_manager.rs` holds the invariant where it is relied on rather than trusting every caller, in one no-op-everywhere helper with a test.

## What I checked

| | |
|---|---|
| svelte-check | 0 errors |
| vitest | 1515 passed |
| Playwright | 50 passed |
| Rust suite | 723 passed, 0 failed |
| `src/ipc/bindings.ts` | regenerated by `cargo test export_bindings` — identical to what is committed |

47 new unit tests cover the platform sniffer, the path join, the menu controller and the two shared item builders, plus the reworked Workdir menu tests — including the two that pin the staged-row and deleted-file rules from your review. Four Playwright specs cover opening a file menu and a folder menu, edge-flipping, and dismissal in a real browser — the folder one also pins that a right-click does not toggle the `<details>` open. `file_manager.rs` has eight, including the traversal shapes the join refuses and the empty relative the repo chip relies on.

### DIFF
--- a/e2e/context-menu.spec.ts
+++ b/e2e/context-menu.spec.ts
@@ -114,6 +114,10 @@ test("right-clicking a folder row opens its own menu, with no leading divider", 
   const folder = page.locator("#tree details.dir > summary").first();
   await expect(folder).toBeVisible();
 
+  // Captured BEFORE the right-click: reading it afterwards and comparing it
+  // to itself is an assertion that cannot fail.
+  const openBefore = await folder.evaluate((el) => (el.parentElement as HTMLDetailsElement).open);
+
   await folder.click({ button: "right" });
 
   const menu = page.locator(".ctxmenu");
@@ -128,10 +132,10 @@ test("right-clicking a folder row opens its own menu, with no leading divider", 
   // A right-click must not toggle the folder open or closed — <details>
   // reacts to a plain click, and the handler has to preventDefault so the
   // native menu does not appear either.
-  const openState = await folder.evaluate((el) => (el.parentElement as HTMLDetailsElement).open);
+  expect(await folder.evaluate((el) => (el.parentElement as HTMLDetailsElement).open)).toBe(openBefore);
+
   await page.keyboard.press("Escape");
   await expect(menu).toHaveCount(0);
-  expect(await folder.evaluate((el) => (el.parentElement as HTMLDetailsElement).open)).toBe(openState);
 
   expect(errors).toEqual([]);
 });
--- a/e2e/context-menu.spec.ts
+++ b/e2e/context-menu.spec.ts
@@ -22,6 +22,9 @@ async function skipWizard(page: import("@playwright/test").Page) {
 // its first file row.
 async function fileRows(page: import("@playwright/test").Page) {
   await page.locator('#refLocal [data-branch="main"]').click();
+  // The commit's files live under the detail panel's "Changes" tab, which is
+  // not the one a commit opens on.
+  await page.locator("#detail .d-tabs .d-tab").nth(1).click();
   // #tree, not "#detail .file": the same file-tree snippet is rendered twice,
   // once inline and once inside the expanded-diff modal, so an unscoped
   // selector matches every row twice.
@@ -104,6 +107,7 @@ test("right-clicking a folder row opens its own menu, with no leading divider", 
   await page.goto("/");
   await skipWizard(page);
   await page.locator('#refLocal [data-branch="main"]').click();
+  await page.locator("#detail .d-tabs .d-tab").nth(1).click(); // the "Changes" tab
 
   // #tree for the same reason fileRows() scopes to it: the tree snippet is
   // rendered twice, inline and inside the expanded-diff modal.
--- a/e2e/context-menu.spec.ts
+++ b/e2e/context-menu.spec.ts
@@ -1,0 +1,133 @@
+import { test, expect } from "@playwright/test";
+
+// Design mode (plain browser, no Tauri mock) — the same base `test` as
+// sidebar-jump-design/scroll-blit. The context menu itself is frontend-only:
+// building the item list, positioning it, and dismissing it never touch the
+// backend, so this covers the whole of what a browser can reach. The two
+// items that DO call Tauri (reveal / open in file manager) are unit-tested
+// against a mocked `commands` instead — there is no backend here to reveal
+// anything with.
+
+// In design mode the setup wizard opens at boot unconditionally and its scrim
+// swallows every click underneath (see sidebar-jump-design.spec.ts, which
+// documents the same dance). Escape is the wizard's own "Skip".
+async function skipWizard(page: import("@playwright/test").Page) {
+  const wizard = page.locator("#setupWizardScrim");
+  await expect(wizard).toHaveClass(/\bon\b/);
+  await page.keyboard.press("Escape");
+  await expect(wizard).not.toHaveClass(/\bon\b/);
+}
+
+// Select a commit so the detail panel's file tree is populated, then hand back
+// its first file row.
+async function fileRows(page: import("@playwright/test").Page) {
+  await page.locator('#refLocal [data-branch="main"]').click();
+  // #tree, not "#detail .file": the same file-tree snippet is rendered twice,
+  // once inline and once inside the expanded-diff modal, so an unscoped
+  // selector matches every row twice.
+  const rows = page.locator("#tree .file");
+  await expect(rows.first()).toBeVisible();
+  return rows;
+}
+
+test("right-clicking a commit's file row opens a menu, and dismisses like a popover", async ({ page }) => {
+  const errors: string[] = [];
+  page.on("pageerror", (e) => errors.push(String(e)));
+  await page.goto("/");
+  await skipWizard(page);
+  const file = (await fileRows(page)).first();
+
+  await file.click({ button: "right" });
+
+  const menu = page.locator(".ctxmenu");
+  await expect(menu).toBeVisible();
+  // Three of the row's own actions, a divider, then the three this menu adds.
+  await expect(menu.locator(".ctxmenu-item")).toHaveCount(6);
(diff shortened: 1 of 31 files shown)


zangjiucheng (changes requested):
Thanks for this. The shared island is the right call, and going through Rust rather than `plugin-opener`'s `allow-open-path` is exactly right given the webview runs plugin code.

One thing to fix before merge. `openRowMenu` (`workdir.svelte.ts:853`) emits the discard item unconditionally, and this PR is what newly attaches the menu to staged rows (`Workdir.svelte:625`; dev has no handler there). On a staged row `untracked` is always false, so `discard_file` takes the worktree-scoped tracked branch: for a file staged with no unstaged edits, `backup_tracked_patch` finds no index-to-workdir delta and returns Err (`workdir.rs:947`), so the user types the filename to confirm and gets an error. For a file in both lists, `git restore --worktree` throws away the unstaged edits and leaves the staged change alone. It also breaks the rule from your own description, since staged rows have no trash button (that lives in `unstagedDirNode`).

```ts
// workdir.svelte.ts, openRowMenu
...(staged
  ? []
  : [{ id: "discard", label: t("workdir.discard_changes"), danger: true, run: () => this.doDiscard(file, untracked) }]),
```

`workdir.svelte.test.ts:563` then needs to expect `["unstage", "reveal", "copy-path", "copy-full-path"]` for a staged row.

Two non-blocking follow-ups. Thread `onDisk` through the working-tree menu, since `status !== "D"` is right there at the call site. And the three backend calls `void` their `Result`, so right-clicking a file from an old commit that no longer exists on disk does nothing at all instead of warning:

```ts
const r = await commands.revealPathInFileManager(repo, relative);
if (r.status === "error") bridge.tama.warn(be(r.error) || t("workdir.reveal_failed"));
```

Also `e2e/context-menu.spec.ts:127` reads `openState` after the right-click, so that assertion can't fail.

Separately: the menu sits at z-index 61 and `.scrim` at 300, so a right-click inside the expanded diff opens the menu behind it. Your CSS change is a pure rename and this reproduces on dev, so it is not yours to fix here. I'll file it.

This needs a rebase too, the branch carries the #98 through #109 stack and is 13 commits behind dev. It lands after #113.

JerryHyun:
Thank you for the careful read — the discard one was a real bug, and you had it exactly right.

## Discard on staged rows

Fixed. It only shows on unstaged rows now.

Your reading of why was right on every point. On a staged row `untracked` is always false, so discard went down the tracked path and acted on the working tree. If the file was only staged, there was no change on disk to back up, so it failed — after the user had already typed the filename to confirm. If the file was in both lists, it threw away the edits on disk and left the staged change sitting there. Neither is what "discard" means to someone looking at that row.

And you're right that it broke this PR's own rule: staged rows have no trash button, so the menu was offering something the row itself doesn't.

`workdir.svelte.test.ts` expects `["unstage", "reveal", "copy-path", "copy-full-path"]` for a staged row now. I also added a test that just checks discard isn't there, so it can't quietly come back.

## The three smaller ones

**`onDisk`.** Threaded through. Both call sites pass `f.status !== "D"`, which was sitting right there as you said. A deleted row now has reveal greyed out, matching the commit view. Both copy items stay enabled — a deleted file's path is often the thing you wanted.

**The ignored results.** All three go through a small shared helper now that warns through Tama. I put it in one place because the open half has two callers. Two new strings, in all three locales.

**The assertion that couldn't fail.** This one was worse than it looked — the PR description pointed at that test as proof the right-click doesn't toggle the folder, and the test wasn't checking anything. It reads the open state before the right-click now. I also checked it the other way: I made a right-click toggle the element on purpose, and the test went red. Before, it would have passed.

## The z-index one

Left it, as you suggested. It happens on `dev` too and the CSS here is only a rename.

## Rebase

Done now that #113 is in, so the stack is out of the diff.

That turned up two specs written against the old layout, fixed in the same commit. `context-menu.spec.ts` looked for a commit's files right after selecting it, but they sit behind the Changes tab now. And `staging.spec.ts` clicked `.wd-rowmenu-item.danger`, the working tree's old popover, which this PR replaces with the shared menu. Tests only — the staging one still goes through the real confirm box and still checks `discard_file` arrives with the right arguments.

## One more thing

Your note on #113 about writing down the alternatives that didn't work sent me back through this branch's comments, and I found one that was wrong. `diritems.ts` said both folder trees "already have their own actions above" the path items — true for the file menu it was adapted from, but a commit's folder row has nothing above them. `ContextMenu.svelte` names that exact case as the reason it drops a leading divider, so the two comments disagreed. Fixed, and I added the reason the shortcut through `plugin-opener` doesn't work to the frontend file, where someone would actually go looking before trying it.

svelte-check 0 errors · vitest 1515 · Playwright 50 · Rust 723 passed.

zangjiucheng (approved):
Verified rather than taken on trust: reverting the staged-row guard turns both tests red, and the staging e2e still drives the real typed-confirm scrim and asserts `discard_file`'s arguments, so nothing was loosened to make this pass. Leaving `paths.ts`'s backslash sniff as a follow-up — it only reaches the clipboard. Thanks for the thorough turnaround.

PR #112, opened 2026-08-25
fix: Do not auto-run plugin panel command-output on open

### LINKED ISSUE #61: Opening a plugin panel can mutate your repository with no click
Opening a plugin panel fires every `command-output` widget it declares, with no user action.

`pluginpanels.svelte.ts:186-188` runs them on panel open, which reaches `if mutates { snapshot_before_mutation(&cwd)? }` at `plugin_exec.rs:489-491`.

So merely *looking at* a panel can write to the repository.

### Fix in the frontend, not the validator

This matters, and it is the opposite of the intuitive order.

`load_from` (`plugin_registry.rs:372-397`) deserializes `plugins.json` and, on any parse error, renames it aside and returns an empty list. **It never calls `validate_manifest`** — that only runs on the install path, at `:604`.

A validator-only rule would therefore protect future installs and leave **every already-installed panel plugin auto-mutating**. Require a click in the frontend first; add the validator rule afterwards as defense in depth.

### Good first issue

Self-contained, one file, and the fix is small. Reading the two code paths above is most of the work.

---

VedantMadane:
## Summary

Do not auto-run plugin panel command-output on open

## Changes

- Opening a panel no longer executes command-output widgets
- Seed idle output slots; user must click re-run to execute (avoids silent repo mutation)

Fixes #61


zangjiucheng (changes requested):
LGTM, it's a good fix, but pls delete [COMMITMSG](https://github.com/zangjiucheng/GitCat/pull/112/changes#diff-1e6192dc1f8b6dc94e9e0f96346df7cd0ae90dbdd3733be00d473ba208327011) and [PRBODY.md](https://github.com/zangjiucheng/GitCat/pull/112/changes#diff-45830d5d219f803137bfdaefaeff739937053a590edd6badbecb09eab9a7e76d) which is garbage file here. Then it's good to merge

VedantMadane:
Removed COMMITMSG and PRBODY.md as requested. The PR now contains only the intended plugin panel fix and tests. Thank you!

zangjiucheng (approved):
LGTM, thanks for your contribution for GitCat!!

PR #113, opened 2026-08-27
feat: put the detail panel along the bottom, and give it tabs

JerryHyun:
You can now put the detail panel along the bottom of the window instead of down the right side, and its contents are split into tabs.

The default doesn't change — the panel still opens on the right, same as always.

### On the right -  the default
right_1
right_2

---

### Along the bottom
bottom_1
bottom_2

## Why I wanted this

Everything about a commit shares one 344px column: the subject, the message, who wrote it and who committed it, the refs, the changed files, and the diff. That's more than fits comfortably in one narrow strip, and it's what got me started on this.

The full-screen diff popup already helps — its comment says as much, that it's there for when a real changeset needs more than the 320px `.diffview` gets. This just adds another way to get that room, without having to open anything.

## What you get

**A bottom placement.** Settings → Graph → **Detail panel** → *Bottom*. The panel goes full width, and the graph gives up height instead of width. Drag its top edge to resize, and it remembers where you left it.

**Tabs, in both placements:**

| | Tabs |
|---|---|
| A commit | **Commit** (message, author, refs) · **Changes** (files and their diff) |
| Uncommitted changes | **Commit** (message box) · **Changes** (staged, unstaged, diff) · **Stash** |

The message and the diff each get the whole panel now, instead of sharing it. In the **Changes** tab the file list sits next to its diff with a divider you can drag — side by side at the bottom, stacked on the right.

I put the tabs in both placements rather than only the new one, because they help the right side too, and you get that without moving anything.

**The stash is one click away** instead of the last thing you scroll to.

## A few notes

- The setting defaults to `right`, so if you never open Settings the panel is exactly where it's always been. The tabs are the only thing that changes there.
- **On the default:** I think a git GUI is more comfortable to work in with the detail panel along the bottom. Would you consider shipping that as the default? It's one line in `DEFAULTS`, so it's easy either way.
- No backend changes at all. Nothing under `src-tauri/`, and `src/ipc/bindings.ts` is untouched.
- This leaves the app with one *fewer* resize implementation, not one more. The two-pane divider was written out inline in two places; now it's one component with four callers. Your panel-edge handles are a separate thing and I've left them alone — folding those together felt like its own change, not something to slip in here.
- The panel-edge drag still only commits its size when you let go, moving a guide line while you drag, since the graph canvas redraws on every size change. Seemed worth keeping, so the new vertical handle uses that same code rather than its own.
- New strings are in `en`, `ko` and `zh`.

## What I checked

| | before | after |
|---|---|---|
| `svelte-check` | 0 errors | 0 errors |
| unit tests | 1440 | **1469** |
| browser tests | 21 | **46** |

The new browser tests run against both placements — the panel moves, the split follows the placement, the divider drags, tabs switch, each view keeps its own tab, and a saved placement survives a reload.

Because the working-tree panel is where a reorganisation like this could do real damage, ten of them cover staging specifically, five scenarios against each placement: stage a file and unstage it, stage a whole folder, discard a file (typing the filename into the confirm box, as the app requires), write a message and commit, and save a stash. Each one drives the actual controls and then checks the backend was asked to do the right thing — `stage_file` with that path, `commit` with that message, and so on — rather than just that a button could be clicked.


zangjiucheng (approved):
Approving. Sorry for the slow review, especially on something this size.

This is well done, and the parts that are easy to get wrong are the parts you got right. Deriving `changesSplitAxis` from the `settingsCtrl` `$state` field rather than the `data-detail-placement` attribute is the correct call, and the unit test that flips placement three times on a live instance is what makes that provable rather than asserted. Same for moving the Escape handler into `DetailPanel.svelte` with a matching e2e regression, and for the peer-handle mechanism reading `collapsed` off the shared `.detail.collapsed` class instead of a private mirror. The staging spec closes a gap that predates this PR entirely, since nothing anywhere staged a file before.

I also want to call out the commenting style, here and in #98: recording the rejected alternatives and the exact error each one produced, in the code rather than in the PR body. That is a norm I want more of in this repo.

One design call I would rather have your read on than inherit: the working-tree view opens on Commit, which is an empty message box, so reaching the staged/unstaged trees costs a click every time and nothing auto-switches the tab. Defaulting `worktreeTab` to `changes`, or reordering to Changes/Commit/Stash, is a one-liner plus the positional `tabs.nth()` indices in the e2e specs. What do you think? I would keep `right` as the default placement for now either way, and revisit once the bottom layout has had real use.

Follow-ups, none blocking:

- The resize guide leaks state across axes: `onUp` at `src/legacy/main.ts:2078` calls `resizeGuide()` with no argument, so the orientation class is stripped mid-fade and neither the inline `top` nor `left` is cleared when the axis flips. After a bottom drag, the sidebar handle's guide line starts at the previous drag's Y. Passing `vert` through and clearing the off-axis property inside `resizeGuide()` fixes it.
- Worth eyeballing the commit-to-commit transition in both placements: `.detail` is now a flex column with `.d-view` at `flex:1`, and the `{#key c.sha}` fade keeps both blocks in flow for 120ms, which should halve the visible content height on every commit click.
- Small ones: `#resizeDetailBottom` has no `title` attribute, `setCollapsed` does not propagate the title to peer handles, and the comment at `Settings.svelte:147` says `settings.detail_placement` already shipped in all three locales when this PR is what adds it.

Merge order: this goes in ~~after #112~~ and before #110, since #110's e2e spec queries `#tree` without clicking the Changes tab your restructure puts it behind.

Issue #83, opened 2026-08-20
Lazy-mount BisectDrawer and PluginPanel


zangjiucheng:
The remaining two modal islands, separated because neither mounts into `document.body`:

- **BisectDrawer** mounts into `#bisectPanelMount` inside `#canvasWrap`, and opens via `openBisectEntry()`
- **PluginPanel** is driven by the plugin panel registry rather than a fixed call site

Both need the lazy-mount pattern from the plain-modal issue, adapted to their own mount points.

**Pattern established in** #81.

---
<sub>Found during v1.3.0 research: parallel codebase surveys, then each finding independently verified against the code before filing. Line numbers are from `dev` at the time of writing.</sub>

Issue #82, opened 2026-08-20
Lazy-mount Terminal — removes 334 kB of xterm from the boot chunk


zangjiucheng:
`Terminal.svelte:6-8` statically imports xterm, and `:47-51` constructs and `open()`s it in `onMount`. So xterm's **334 kB** is in the boot chunk and its terminal is constructed at startup, for every user, whether or not they ever open a terminal.

### The trap that makes this its own issue

Terminal must be **mount-once-on-first-open and then stay mounted**. `main.ts:243-249` explains the scrollback keep-alive: unmounting on close throws away the user's terminal history.

So this is not the same mechanical change as the other modals, and it should not be reviewed as if it were.

**Pattern established in** #81.

---
<sub>Found during v1.3.0 research: parallel codebase surveys, then each finding independently verified against the code before filing. Line numbers are from `dev` at the time of writing.</sub>

PR #124, opened 2026-09-03
perf: mount the four plain modal islands on first open

### LINKED ISSUE #81: Lazy-mount the four plain modal islands
`src/main.ts` has 72 static imports, 37 top-level `mount()` calls and **zero** dynamic `import()`. Every modal island is downloaded, parsed and mounted into `document.body` before the user touches anything.

Measured for the full modal batch: **214 → 122 ms DCL**, **1151 → 889 DOM nodes**. Islands account for 131 ms of the 214.

### Scope: the four with no special mounting

`TamaGallery`, `Plugins`, `Settings`, `ExternalTools` — convert to first-use `() => import(...)` thunks. This establishes the pattern; Terminal and the non-body mounts follow in their own issues.

**Keep the controllers eager.** They are used at boot (`main.ts:426, 462, 524, 599, 632, 635, 638, 656, 716`). Only the views lazify.

### Not in scope

`manualChunks` was measured and rejected: `@lucide/svelte` is 7.44 kB and `@tauri-apps/*` 5.95 kB, 0.75% combined, because lucide already imports per-icon. And `src/legacy` is 1,112 kB of the 1.77 MB bundle and is not lazifiable, so "code-split the bundle" is not the framing.

### Measurement

Production build under `vite preview`; DCL and DOM-node count before and after.

**Siblings:** #82 (Terminal) and #83 (non-body mounts). This one establishes the pattern.

---

zangjiucheng:
Closes #81. Establishes the pattern that #82 (Terminal) and #83 (non-body mounts) follow.

## What moved

`TamaGallery`, `Plugins`, `Settings`, `ExternalTools` — the four the issue scoped as "no special mounting". Only the **views** move; the controllers stay eager, because boot code reads them (menu wiring, ⌘K registration, settings applied to the document).

That split is what makes this cheap. `open` is plain `$state` on a singleton that already exists, so `mountOnFirstOpen(() => ctrl.open, () => import(...))` needs no new plumbing on either side.

## What I checked before moving each one

Deferring a mount defers whatever that mount did, so eligibility is not a matter of taste:

- None of the four views has an `onMount`, an `$effect`, an `addEventListener`, or a prop. They are pure presentation.
- `open` is assigned **only** inside each controller's own module — grepped across `src/` — so watching it is a complete trigger, not one entry point among several.
- All four mounted identically: `mount(X, { target: document.body })`, no props.

`lazyisland.svelte.ts` states that bar in its header, because the failure mode for getting it wrong is silent: the work simply stops happening until someone opens the modal.

## Measured, on a production build

| | before | after | delta |
| --- | --- | --- | --- |
| boot JS | 1743.54 kB | 1698.97 kB | **−44.56 kB** (−11.75 kB gzip) |
| boot CSS | 20.34 kB | 16.51 kB | **−3.83 kB** |
| **boot total** | | | **−48.39 kB raw, −12.45 kB gzip** |
| chunks | 5 | 10 | |

The four chunks total 48.87 kB, so this is a *move*, not a saving. The point is that none of it is on the boot path any more.

The issue's headline figures (214 → 122 ms DCL, 1151 → 889 DOM nodes) are for the full modal batch and need a browser to reproduce; I could not run one here, so I am not restating them as if I had.

## Two things worth deciding on rather than inheriting

**The first open waits for its chunk.** Off the local asset protocol that is well under a frame, but it does mean the first open of a modal does not animate its scrim in, because there is no scrim in the DOM yet to animate. Subsequent opens are exactly as before. If that bothers you, a `requestIdleCallback` prefetch after boot would keep the chunk warm without putting it back on the boot path — I left it out as scope.

**A failed chunk load clears the guard and logs** rather than becoming an unhandled rejection. Worth naming because I flagged exactly this shape in #94's locale loader, and it would have been the same mistake here.

## Tests

A new e2e spec asserting the property that actually distinguishes this change: the four are **absent from the DOM** at boot, not merely hidden. `toBeHidden()` passes for both "mounted and hidden" and "not mounted", so it cannot tell the old behaviour from the new one; `toHaveCount(0)` can. It also checks that opening one mounts exactly that one — not its siblings, which is what a single shared chunk would have done — and that close/reopen does not produce a second copy in `document.body`.

The existing `settings-shortcut.spec.ts` is an independent guard that the modal still actually opens.

⚠️ **The new spec has not been run locally** — the Playwright browser download fails here with `ECONNRESET`, so CI is its first execution. `svelte-check` is clean and all 1526 unit tests pass.

### DIFF
--- a/e2e/lazy-islands.spec.ts
+++ b/e2e/lazy-islands.spec.ts
@@ -1,0 +1,70 @@
+import { test, expect } from "@playwright/test";
+
+// #81: the four plain modal islands are mounted on FIRST OPEN, not at boot.
+//
+// The distinction these tests turn on is "absent from the DOM" versus "present
+// and hidden". Every one of these modals was previously mounted into
+// document.body at boot and rendered a hidden scrim; a `toBeHidden()` assertion
+// passes either way, so it cannot tell the two apart. `toHaveCount(0)` can.
+
+const MODALS = [
+  { name: "Settings", selector: ".modal.settings" },
+  { name: "Plugins", selector: ".modal.plugins-modal" },
+  { name: "External Tools", selector: ".modal.external-tools" },
+  { name: "Tama Gallery", selector: ".modal.tama-gallery" },
+];
+
+test("none of the lazily mounted modals is in the DOM at boot", async ({ page }) => {
+  const errors: string[] = [];
+  page.on("pageerror", (e) => errors.push(String(e)));
+
+  await page.goto("/");
+  await page.waitForTimeout(300);
+
+  for (const m of MODALS) {
+    await expect(page.locator(m.selector), `${m.name} must not be mounted before it is opened`).toHaveCount(0);
+  }
+
+  // The controllers, unlike the views, stay eager — boot code reads them. If a
+  // controller had been lazified too, the menu/⌘K wiring that calls into it at
+  // boot would have thrown by now.
+  expect(errors).toEqual([]);
+});
+
+test("opening one mounts exactly that one, and only once", async ({ page }) => {
+  const errors: string[] = [];
+  page.on("console", (m) => {
+    if (m.type() === "error") errors.push(m.text());
+  });
+  page.on("pageerror", (e) => errors.push(String(e)));
+
+  await page.goto("/");
+  await page.waitForTimeout(300);
+
+  const settings = page.locator(".modal.settings");
+  await expect(settings).toHaveCount(0);
+
+  await page.keyboard.press("Control+,");
+  await expect(settings, "Ctrl+, should load and mount Settings").toBeVisible();
+
+  // Its siblings must still be absent — opening one modal must not drag the
+  // other three in with it, which is what a single shared chunk would do.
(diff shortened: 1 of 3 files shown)


zangjiucheng:
Correcting the record on this PR, since I made a claim about the check that was stronger than the check itself.

I wrote that none of the four views has "an `onMount`, an `$effect`, an `addEventListener`, or a prop". All four *do* register a global keydown handler — via `<svelte:window on:keydown>`, which the grep behind that sentence would not have matched. I found it while doing the same eligibility pass for #82/#83, where BisectDrawer and PluginPanel use the same form.

The conclusion still holds, and I checked rather than assumed: all four handlers are `if (e.key === "Escape" && ctrl.open) ctrl.close()`. A listener guarded by the very flag the mount is keyed to cannot matter before the mount, because the mount is what that flag triggers. So nothing regressed here.

What was wrong was the rigour I claimed, not the outcome. `lazyisland.svelte.ts`'s eligibility bar now names `<svelte:window>` explicitly, and says why a listener of this shape is still fine — the follow-up PR carries that change.

PR #146, opened 2026-09-14
docs: document Tama Gallery keyboard exception

Ahiram15:


zangjiucheng:
Thanks for the contribution, but unfortunately, I won't include the Easter egg in the docs.

